PR #19, opened 2026-03-26
ci: run arm reproducible build on native runner

omkhar:
## Summary
- move the CI arm64 reproducible-build lane from x64+QEMU to the native `ubuntu-24.04-arm` hosted runner
- keep amd64 on `ubuntu-latest`
- tighten `scripts/check-pinned-inputs.sh` to pin the exact reviewed matrix structure for `reproducible-build-platform`

## Why
Recent successful CI runs show the arm64 reproducible-build lane is the dominant bottleneck:
- last 10 successful CI runs on March 26, 2026: arm64 averaged about `24m08s`
- the same runs: amd64 averaged about `6m13s`
- arm64 was about `3.89x` slower on average

The current workflow runs both lanes on `ubuntu-latest`, and only the arm64 lane turns on QEMU. This change targets that measured emulation overhead directly.

A bigger GitHub-hosted ARM runner is not a practical option on this repo as it stands: `omkhar/workcell` is a private user-owned repo, and GitHub larger runners are not available here. The realistic lever is native `ubuntu-24.04-arm`, not a bigger box.

## Validation
- `./scripts/check-pinned-inputs.sh`
- `./scripts/check-workflows.sh`
- `./scripts/validate-repo.sh`

## Rollout
If PR CI is green, compare the next few successful runs against the current arm64 baseline before changing anything else in the reproducible-build path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,7 @@ jobs:
 
   reproducible-build-platform:
     name: Reproducible build (${{ matrix.platform_name }})
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
     needs: validate
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


omkhar (commented):
Follow-up local review before pushing the latest signed commit:

- tightened the invariant checker so it now explicitly proves there is exactly one `linux/arm64` matrix entry and that it uses `ubuntu-24.04-arm`
- re-ran `./scripts/check-pinned-inputs.sh`, `./scripts/check-workflows.sh`, and `./scripts/validate-repo.sh` locally on the updated branch

Holding the push until the fresh strict review returns on the exact patch.

omkhar (commented):
Peer review note before update: the current head passed local validation, but I found and fixed a gap in the runner invariant enforcement locally. I am holding merge and pushing the tighter signed checker update next so the reviewed invariant matches the intended native-arm runner policy.

omkhar (commented):
Peer review note: tightened the native arm runner invariant so the CI checker now explicitly requires exactly one `linux/arm64` reproducible-build lane on `ubuntu-24.04-arm`, in addition to the reviewed matrix block check. Local validation on head `5f9ccfd`: `./scripts/check-pinned-inputs.sh`, `./scripts/check-workflows.sh`, `./scripts/validate-repo.sh`.

omkhar (commented):
Runtime/parallel work status while CI runs: the smoke/helper branch is locally green on `bash -n`, `shellcheck`, `./scripts/validate-repo.sh`, and a Docker-backed `WORKCELL_CONTAINER_SMOKE_DOCKER_CONTEXT=colima ./scripts/container-smoke.sh`. I am keeping it local until this PR finishes so the branches do not stack.

PR #81, opened 2026-04-10
Harden development wrapper against protected runtime bypass

omkhar:
### Motivation

- The development wrapper previously `exec`'d arbitrary user commands, and because the exec guard treated that wrapper as approved, protected provider runtimes could be invoked from development mode and bypass the `provider-policy` checks and breakglass protections.
- The change prevents development-mode arbitrary-command handling from being used to directly launch protected runtimes or provider commands while preserving normal development workflows for non-protected binaries.

### Description

- Add `reject_protected_runtime_launch()` to `runtime/container/development-wrapper.sh` to inspect the intended launch target and abort if it matches protected runtime paths or provider command names.
- The guard rejects direct paths such as `/usr/local/libexec/workcell/real/codex`, `/usr/local/libexec/workcell/real/claude`, `/usr/local/libexec/workcell/real/node`, `/usr/local/libexec/workcell/real/git`, `/usr/local/libexec/workcell/core/git`, `/usr/local/libexec/workcell/git`, and `/usr/local/libexec/workcell/node`.
- The guard also rejects provider command basenames `codex`, `claude`, and `gemini`, and is invoked immediately before `exec "$@"` so development-mode commands cannot bypass provider-policy checks via the approved-wrapper context.

### Testing

- Ran a shell syntax check `bash -n runtime/container/development-wrapper.sh`, which completed successfully.
- There are no automated unit tests in the repository covering these wrapper scripts.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d6aa2de400832984b894e8895ef1ba)


omkhar (commented):
Peer review summary:

- Confirmed the original wrapper patch was incomplete because literal-path matching still allowed renamed or symlinked paths to the protected runtimes.
- Verified the fix now uses same-file matching for direct path launches, so renamed symlinks to the protected runtime payloads are rejected before the approved wrapper context can bypass provider-policy.
- Added container smoke coverage for that regression path so the review finding is locked into CI.
- Fixed the Go CodeQL misconfiguration by moving Go to `autobuild` while keeping Rust and JavaScript buildless, and mirrored that release-side so tagged releases do not regress into a broken CodeQL init path.
- Added workflow validation coverage so future edits cannot silently reintroduce `build-mode: none` for Go or remove the release-side CodeQL gating.
- Refreshed the Debian snapshot pins because the branch was blocked on upstream drift at commit time.

Residual risk checked:
- Bare provider command names remain blocked (`codex`, `claude`, `gemini`).
- Managed development-mode non-provider commands are preserved.
- Control-plane manifest parity was regenerated and revalidated after the wrapper change.

omkhar (commented):
Follow-up peer review on commit `6b7c73c`:

- The remaining `Container smoke` failure was a harness bug, not a wrapper regression. The test created `/state/tmp` as root with default `0755`, then the development wrapper re-execed as the mapped runtime user and lost write access. Moving the renamed-runtime fixture to `/workspace/tmp` keeps the regression test on a path that remains writable under the mapped UID model used by the smoke harness.
- The two open CodeQL findings on `main` are addressed by narrowing the escape parser to fixed-width helpers: `\x` now parses at 8 bits and `\u`/`\U` parse at 16/32 bits before conversion, rather than flowing through a 64-bit `ParseUint` result into `byte`/`rune`.
- Added direct parser coverage for valid and invalid `\x`, `\u`, and `\U` escapes so the security fix is behavior-locked in tests instead of relying only on code scanning.

Local verification completed before push:
- `go test ./internal/policybundle ./internal/metadatautil ./internal/testkit`
- `bash -n scripts/container-smoke.sh`
- `./scripts/container-smoke.sh`

Residual risk:
- Waiting on the GitHub rerun for CI, CodeQL, docs, and security on head `6b7c73c` before merge.

omkhar (commented):
Final peer-review update on commit `ab00113`:

- The remaining Linux-only CI failure was in the smoke harness, not the runtime boundary. The renamed-runtime fixture was still mutating the mounted `/workspace` as container root, which works locally on Colima but is denied on GitHub-hosted Docker volumes where the writable path is owned by the mapped runner UID.
- The fix now creates and removes that renamed symlink through `setpriv --reuid "$WORKCELL_HOST_UID" --regid "$WORKCELL_HOST_GID" --init-groups`, matching the rest of the harness's writable-workspace behavior and the runtime's actual mapped-user model.
- Re-ran `./scripts/container-smoke.sh` locally after the patch; the full harness passed, including the renamed-runtime regression path.

Waiting only on the GitHub CI rerun for head `ab00113` before merge.

omkhar (commented):
Final pass on the signed squash commit `0813798d79683322d351418934540838f6269b86` is clean. The tree matches the previously green PR content, CI is green across validation/smoke/repro/install workflows, and the default-branch CodeQL findings should clear once this exact commit lands on `main` and default-branch analysis reruns.

PR #86, opened 2026-04-11
security: fix high-severity findings

therandomsecurityguy:
## Summary

- Fix TOCTOU race in secret file validation (`RequireSecretFile`, `requireSecretFile`) by opening the file first and using `f.Stat()` on the fd instead of separate `Lstat`/`Stat` calls
- Cache protected runtime signatures in `OnceLock` to eliminate 8–16 redundant `stat()` syscalls per `execve` interception
- Add `USER nobody` to all three Dockerfiles as defense-in-depth against direct `docker run` without `--user`; host launcher already passes `--user root` explicitly
- Replace predictable PID-based temp files (`$$`) with `mktemp` in `runtime-user.sh` and `apt-helper.sh`; use atomic `mktemp`+`mv` pattern for assurance file writes

## Findings addressed

| # | Severity | Finding | Fix |
|---|----------|---------|-----|
| H1 | High | Protected runtime signatures recomputed on every `execve` (8–16 stat() syscalls/call) | `OnceLock` caching in `rust/src/lib.rs` |
| H2 | High | Runtime Dockerfile runs as root by default (no `USER`) | `USER nobody` + comment |
| H3 | High | Remote-validator installs docker-cli (host escape risk) | `USER nobody` + WARNING comment |
| H4 | High | TOCTOU in `RequireSecretFile` — Lstat then Stat | `os.OpenFile` + `f.Stat()` on fd |
| H5 | High | TOCTOU in `requireSecretFile` — Lstat then Stat | Same fix as H4 |
| H6 | High | PID-based temp files (`$$`) predictable; TOCTOU in apt-helper writes | `mktemp` + atomic `mv` pattern |

## Test plan

- [x] All Go tests pass (`go test ./...`)
- [x] `go vet ./...` clean
- [x] `shellcheck -x` clean on modified shell scripts
- [x] Rust changes are structurally valid (full build requires container toolchain)


internal/authpolicy/policy.go
```diff
@@ -1050,17 +1050,26 @@ func requireSecretFile(source string, label string) (string, error) {
 	if info.Mode()&os.ModeSymlink != 0 {
 		return "", die(fmt.Sprintf("%s must not be a symlink: %s", label, source))
 	}
-	if !info.Mode().IsRegular() {
+	f, err := os.OpenFile(source, os.O_RDONLY, 0)
```
omkhar, line 1053: I reproduced a deterministic symlink-swap race here by replacing the checked file with a symlink after the `Lstat` precheck but before `os.OpenFile(..., os.O_RDONLY)`. The open followed the swapped symlink and the validation accepted the attacker-controlled target, so the TOCTOU gap remains. Please switch this to a no-follow or fd-relative open pattern; that variant passed the repro locally.

runtime/container/bin/apt-helper.sh
```diff
@@ -77,24 +77,23 @@ workcell_validate_apt_args() {
...
   fi
-  if [[ -e "${WORKCELL_PERSISTED_ASSURANCE_FILE}" ]]; then
-    chmod u+w "${WORKCELL_PERSISTED_ASSURANCE_FILE}" 2>/dev/null || true
-  fi
-  printf '%s\n' "lower-assurance-package-mutation" >"${WORKCELL_PERSISTED_ASSURANCE_FILE}"
-  chmod 0444 "${WORKCELL_PERSISTED_ASSURANCE_FILE}" 2>/dev/null || true
+  tmp_file="$(mktemp "${WORKCELL_PERSISTED_ASSURANCE_FILE}.tmp.XXXXXX")"
+  printf '%s\n' "lower-assurance-package-mutation" >"${tmp_file}"
+  chmod 0444 "${tmp_file}"
+  mv "${tmp_file}" "${WORKCELL_PERSISTED_ASSURANCE_FILE}" 2>/dev/null || true
```
omkhar, line 96: I forced the persisted-marker rename to fail while leaving the runtime marker write intact. With `|| true` here the function still exits successfully and silently drops the persisted assurance file, so the lower-assurance state is no longer durably recorded. This should fail closed instead of suppressing the rename failure.

internal/policybundle/policy_bundle.go
```diff
@@ -925,25 +925,26 @@ func RequireSecretFile(source string, label string) (string, error) {
 	if err := requireNoSymlink(source, label); err != nil {
 		return "", err
 	}
-	info, err := os.Stat(source)
+	f, err := os.OpenFile(source, os.O_RDONLY, 0)
```
omkhar, line 928: I reproduced the same symlink-swap race here: after `requireNoSymlink` returns, swapping `source` to a symlink before `os.OpenFile(..., os.O_RDONLY)` still redirects the validation onto the attacker-controlled target. Please use a no-follow or fd-relative open here as well; that change closed the repro locally.

omkhar:
Two additional follow-ups on the same TOCTOU class, both still within the scope of this patch set:

1. `internal/authpolicy/manage.go`
`requireSecretFile` is only the validation step. The actual copy path still re-opens the same pathname in `writeSourceFile`, so a symlink swap after validation but before the copy still wins. I reproduced this deterministically: validate a good file, replace it with a symlink, then the managed copy consumes the attacker target. Please either carry a verified fd through the call chain or re-open the source with no-follow/fd-relative semantics at use time, not just in the validator.

2. `internal/authresolve/resolve_credential_sources.go`
`resolveClaudeMacosKeychain` has the same pattern: `requireSecretFile` validates the export path, then `materializeFileUnderRoot` later re-opens it by name. That leaves the same swap window in the resolver-backed path. Please apply the same no-follow/fd-relative reopen strategy here as well.

omkhar:
Superseded by the split follow-up work:

- #89 `security: harden secret-file opens and assurance writes` (merged)
- #90 `runtime: cache protected runtime signatures in exec guard` (merged)

Those PRs preserve Derek Chamorro's authored commits on `main` via `git cherry-pick -x`, so contributor credit is retained in the project history.

I intentionally did **not** merge the `USER nobody` Dockerfile slice from #86 yet. That hardening change still needs a follow-up to keep the validator/direct-script bootstrap path working under an unprivileged default user, and I did not want to mix that unresolved runtime-defaults rollout into the already-reviewed security fixes above.

PR #222, opened 2026-05-13
^R Translate session_logs_main into internal/sessionctl/LogsMain

omkhar:
## Summary

PR 22.2 in the /sethify-remediation chain (Section B mid, sibling of #219 / PR 22.1). Mirrors that bootstrap pattern: bash `session_logs_main` becomes a one-line shim that delegates to `go_hostutil launcher session-logs-cli`, and the substantive work moves into `internal/sessionctl/LogsMain`.

Behavior preserved (with one deliberate exception, below):

| Bash surface | Go equivalent |
|---|---|
| `--id` / `--kind` parsing + missing-arg errors | `parseLogsArgs` returning the bash error strings verbatim |
| `validate_logs_kind_name` | `validateLogsKindName` — emits the legacy two-line `Unsupported log kind: ...\nUse --logs audit, --logs debug, --logs file-trace, or --logs transcript.` |
| `load_session_runtime_metadata` + log-path switch | `sessions.FindSessionRecordInRoots` + `logPathForKind(record, kind)` |
| `revalidate_recorded_host_output_path` | `hoststate.ResolveHostOutputCandidate` |
| Empty path OR missing file | `No <kind> log is recorded for session <id>.` (matches bash byte-for-byte) |
| Revalidation mismatch | `Workcell blocked host output path after launch: <path>` |

The one deviation: exit codes change from bash's `exit 2` to Go's default `exit 1` on parse / validation failures. `verify-invariants.sh:6500` only asserts the stderr message and a non-zero exit, so the existing tests still pass. This matches the deviation PR 22.1 (#219) already established for `session-timeline`.

`scripts/workcell`: **9221 → 9166 LOC**. `control-plane-manifest.json` regenerated.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/sessionctl/ ./cmd/workcell-hostutil/`
- [x] `bash -n scripts/workcell` clean
- [x] `gofmt -l .` empty
- [x] `./scripts/workcell --agent codex --dry-run` smoke: exit 0
- [x] `scripts/check-pr-shape.sh`: files=5 lines=312 areas=4 binary_files=0
- [ ] CI: macOS `Validate repository` exercises `workcell session logs --kind transcript` against a recorded live detached session and asserts the "No transcript log is recorded" message (verify-invariants.sh:6500) — real gate

### DIFF
--- a/cmd/workcell-hostutil/main.go
+++ b/cmd/workcell-hostutil/main.go
@@ -135,6 +135,7 @@ func launcherSubcommands() []launcherSubcommand {
 	return []launcherSubcommand{
 		{"session-usage", 0, 0, cmdLauncherSessionUsage},
 		{"session-timeline-cli", 0, -1, cmdLauncherSessionTimelineCli},
+		{"session-logs-cli", 0, -1, cmdLauncherSessionLogsCli},
 		{"session-suffix", 0, 0, cmdLauncherSessionSuffix},
 		{"colima-status", 1, 1, cmdLauncherColimaStatus},
 		{"cleanup-stale-log-pointers", 1, 1, cmdLauncherCleanupStaleLogPointers},
@@ -194,6 +195,10 @@ func cmdLauncherSessionUsage(_ []string) error {
 
 func cmdLauncherSessionTimelineCli(args []string) error {
 	return sessionctl.TimelineMain(args)
+}
+
+func cmdLauncherSessionLogsCli(args []string) error {
+	return sessionctl.LogsMain(args)
 }
 
 func cmdLauncherSessionSuffix(_ []string) error {
--- a/internal/sessionctl/logs.go
+++ b/internal/sessionctl/logs.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/omkhar/workcell/internal/host/hoststate"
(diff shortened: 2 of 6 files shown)


omkhar:
Pushed a fixup (790bfe5) addressing the macOS Validate lane regression.

The first commit translated `session_logs_main` but the resulting Go `LogsMain` used `os.Getenv("WORKCELL_STATE_ROOT")` / `os.Getenv("COLIMA_STATE_ROOT")` for state-root lookup. `go_hostutil` invokes the binary through `env -i …`, so those vars are scrubbed and `FindSessionRecordInRoots` returned `os.ErrNotExist` ("file does not exist") even when the record existed. `tests/scenarios/shared/test-session-commands.sh:405-408` (the tampered debug-symlink assertion) tripped on this.

The fixup brings the shim/Go contract in line with `session-{show,list,runtime-metadata}` — bash prepends `--root=PATH` from `session_lookup_root_args`, and `LogsMain` / `TimelineMain` consume those args before subcommand parsing via a shared `consumeRootArgs` helper. The same env-only bug existed in `TimelineMain` from PR 22.1 (#219); it was dormant because `session_logs` failed first under `set -e`. Fixed in the same commit since the test now exercises both paths.

`tests/scenarios/shared/test-session-commands.sh` runs to completion locally (`exit 0`).

PR #305, opened 2026-05-30
Build the arm64 release image on a native runner

omkhar:
## Summary

Replace QEMU emulation for the `linux/arm64` release image with native
`ubuntu-24.04-arm` runners, completing the CI→release native-arm migration the
maintainer already scaffolded (the pinned-inputs policy already forbids QEMU in
`ci.yml`; this extends the same treatment to `release.yml`).

QEMU arm64 emulation is the long pole of the release build (~5–10× slower than
native and the slowest step on every release run); native runners are GA and
free for public repos.

## What changed

- **`preflight-arm64-repro`** (new, native arm64, ungated): reproducibility-verifies
  the arm64 runtime image via `verify-reproducible-build.sh` and exports its
  image/config digests as job outputs. `preflight` now builds amd64 natively and
  merges the arm64 digests into the combined `workcell-runtime-preflight.json`.
- **`build-arm64-image`** (new, native arm64, gated on the `release` environment):
  builds and pushes the arm64 image by digest; `release` consumes that digest to
  assemble the multi-arch manifest. Gating preserves the invariant that **no
  image reaches GHCR before the release approval**.
- Drop `docker/setup-qemu-action` from both `preflight` and `release` (the amd64
  lanes were already native).
- **pinned-inputs policy**: replace the now-obsolete `cache-image: false` QEMU
  rule with a rule rejecting `docker/setup-qemu-action` in `release.yml` and
  requiring a native `ubuntu-24.04-arm` lane — mirroring the existing `ci.yml`
  rule.
- **workflow-lane policy**: register the two new release-only lanes and
  regenerate `policy/workflow-lanes.json`.
- Document the native arm64 release lanes.

## Correctness / why this is safe

- The cross-job per-platform digest match reuses the **exact equivalence already
  proven for amd64** today (preflight builds via `verify-reproducible-build.sh`
  OCI export with `provenance=false,sbom=false`; publish builds via
  `build-push-action` image export with `provenance=mode=max,sbom=true`; the
  per-platform `image_manifest_digest`/`config_digest` already match across that
  difference — attestation manifests are filtered out).
- Native arm64 reproducibility is **already proven in `ci.yml`** (the
  `reproducible-build-platform` matrix builds the same image twice on
  `ubuntu-24.04-arm` and asserts identical digests).
- Both the arm64 reproducibility check and the arm64 publish build use the
  **same checked-out context**, so the expected (preflight) and actual
  (published) arm64 digests are produced identically.
- The change is **fail-closed**: the existing *"Verify published platform
  digests match preflight"* gate aborts the release on any mismatch.

## Testing

`release.yml` only runs on tag push, so the end-to-end digest match is exercised
on the **first tagged release after merge** (fail-closed). Everything else was
validated locally:

- `go build/vet/test ./...`
- `actionlint` (all workflows), `zizmor --persona auditor`
- `scripts/check-pinned-inputs.sh` (real workflows)
- `scripts/verify-workflow-lanes.sh`
- `scripts/pre-merge.sh --profile pr-parity`
- `codespell`, `markdownlint`

## Note for the first release

On the first release tag after this merges, the `release` environment may
surface **two** pending-deployment approvals (the new `build-arm64-image` job
and `release`), since the arm64 build is gated to keep GHCR pushes behind the
release approval.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,10 +71,60 @@ jobs:
       - if: ${{ !github.event.repository.private || vars.WORKCELL_ENABLE_PRIVATE_CODE_SCANNING == 'true' }}
         uses: github/codeql-action/analyze@9e0d7b8d25671d64c341c19c0152d693099fb5ba # v4.35.5
 
+  preflight-arm64-repro:
+    name: Release reproducible image (arm64)
+    # Build and reproducibility-verify the linux/arm64 runtime image on a
+    # native arm64 runner instead of QEMU emulation.  The amd64 lane stays in
+    # the main `preflight` job (native on ubuntu-latest); this job feeds its
+    # verified arm64 image/config digests back via job outputs so preflight can
+    # assemble the combined runtime manifest that publish checks against.  This
+    # mirrors the already-reviewed native arm64 lane in ci.yml.
+    runs-on: ubuntu-24.04-arm
+    timeout-minutes: 45
+    permissions:
+      contents: read
+    outputs:
+      image_manifest_digest: ${{ steps.repro.outputs.image_manifest_digest }}
+      config_digest: ${{ steps.repro.outputs.config_digest }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - id: source_date
+        name: Compute source date epoch
+        run: echo "epoch=$(git log -1 --pretty=%ct "${GITHUB_SHA}")" >> "${GITHUB_OUTPUT}"
+
+      - uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # v4.0.0
+        with:
+          cache-binary: false
+          driver-opts: image=${{ env.WORKCELL_BUILDKIT_IMAGE }}
+          version: ${{ env.WORKCELL_BUILDX_VERSION }}
+
+      - id: repro
+        name: Verify reproducible arm64 runtime image
+        env:
+          WORKCELL_REPRO_PLATFORMS: linux/arm64
+          WORKCELL_REPRO_MANIFEST_PATH: dist/workcell-runtime-arm64.json
+          SOURCE_DATE_EPOCH: ${{ steps.source_date.outputs.epoch }}
+        run: |
+          ./scripts/verify-reproducible-build.sh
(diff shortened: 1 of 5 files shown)


omkhar:
## Adversarial review — clean

Independent adversarial review (8 scrutiny dimensions) + re-ran the repo gates (`go test ./...`, `go vet`, `check-pinned-inputs.sh`, `verify-workflow-lanes.sh`) on the committed tree. **No correctness or security problem found.** Summary:

1. **Dependency graph** — OK. `preflight needs preflight-arm64-repro` (merge step can't run before arm64 digests exist); `release needs build-arm64-image`. A hard `needs` edge means `release` is skipped, never run with an empty input. If a digest were ever empty, the `Verify published platform digests match preflight` step (`set -euo pipefail`) fails closed.
2. **jq merge** — OK. Key-assignment preserves the existing `linux/amd64` entry + `source_date_epoch`; digests passed via `--arg`; pre-validated `sha256:*`.
3. **Cross-exporter digest equivalence** — OK. arm64 reuses the *exact* preflight(OCI, provenance/sbom off) vs publish(image, provenance=max/sbom) boundary the amd64 path already relies on; attestation manifests are filtered before the per-platform digest compare.
4. **Build context** — OK. arm64 actual (build-arm64-image, context `.`) vs arm64 expected (preflight-arm64-repro, also context `.`) use identical context — tighter than amd64, which already crosses `.`↔release-source and passes. `.dockerignore` restricts context to tracked source regardless of root.
5. **QEMU removal** — OK, no orphans. `WORKCELL_QEMU_IMAGE` still defined in both ci/release.yml (matching policy intact) and still recorded by the builder-environment manifest; only the `setup-qemu-action` steps were removed.
6. **pinned-inputs policy** — OK. New rules satisfied by the actual workflow; no other QEMU/cache-image assertion left dangling. Mirrors the existing ci.yml treatment.
7. **Approval gating / least-privilege** — OK. `build-arm64-image` gated on `environment: release` (same as publish) → no GHCR push before approval; permissions are `contents: read` + `packages: write` only (correctly omits id-token/attestations/contents:write). Untagged push-by-digest is the same mechanism the prior in-job build and the amd64 path use — no new supply-chain exposure.
8. **Ordering/refs/timeouts/needs** — OK. `imagetools create` amd64→arm64 order and the `["amd64","arm64"]` assertion unchanged.

**Residual:** the cross-job arm64 digest match is only exercised on the first real tag (fail-closed). Gating correctness also depends on the `release` Environment having its required-reviewer rule — that lives in repo settings, identical to the pre-existing publish job, so no *new* gating risk is introduced here.

PR #307, opened 2026-05-30
CI: persist validator Go/Cargo build cache across runs

omkhar:
## Summary

Optimization #2 of the CI/deploy series. The `Validate repository` job (~11m, the CI gating critical path after PR #306) recompiles Go + Rust from cold every run because the validator container's cache dirs are ephemeral. This persists them across runs.

- `ci.yml`: restore a host dir via `actions/cache` (SHA-pinned v5.0.5), keyed on `go.sum` + `runtime/container/rust/Cargo.lock`, passed to the validate step as `WORKCELL_VALIDATOR_HOST_CACHE_DIR`.
- `run-validate-in-validator.sh`: bind-mount that dir over `${validator_cache}` **only when the env var is set**; the default path is byte-identical to today (other callers unaffected).

## ⚠️ Empirical validation in progress

Cache savings only appear on a **warm** run. This PR's first CI run is **cold** (populates the cache, slightly slower from the save). I'll trigger a second run to measure warm savings and post the cold-vs-warm numbers here. **If warm savings aren't meaningful net of cache I/O, I'll close this PR** — per the repo's "empirical evidence, not guesses" rule.

## Invariant safety

- The four `-e GOCACHE/GOMODCACHE/CARGO_TARGET_DIR/XDG_CACHE_HOME` lines + `mkdir` that `verify-invariants.sh` asserts are kept **byte-identical** (verified locally).
- Reproducible-build job is separate and stays `--no-cache` + double-build.
- GitHub scopes caches per-branch (a PR cannot poison main's cache); Go cache is content-addressed, Cargo fingerprinted — a stale entry rebuilds, never mis-serves a pass.
- Cached artifacts are build/test intermediates, never shipped.

## Posture note

This introduces caching into the **validation** pipeline (previously clean-from-cold). It's correctness-preserving, but it is a posture change worth your explicit call — flagging for review rather than auto-merging.

## Testing

Local: `verify-invariants.sh`, `actionlint`, `zizmor` (auditor), `shellcheck`, `check-pinned-inputs.sh`, `verify-workflow-lanes.sh`, `go test ./internal/metadatautil/...`, full `pre-merge.sh --profile pr-parity` (exercises the default unmounted path).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,9 +84,25 @@ jobs:
           driver-opts: image=${{ env.WORKCELL_BUILDKIT_IMAGE }}
           version: ${{ env.WORKCELL_BUILDX_VERSION }}
 
+      # Persist the validator's Go/Cargo build+test artifacts across runs to cut
+      # the from-cold compile cost inside the validate container.  Keyed on the
+      # dependency lockfiles so a toolchain/dep change is a clean miss.  GitHub
+      # scopes caches per-branch (a PR cannot poison main's cache), and the
+      # cached artifacts are content-addressed (Go) / fingerprinted (Cargo), so
+      # a stale entry forces a rebuild, never a wrong pass.  The
+      # reproducible-build job is separate and stays --no-cache.
+      - name: Restore validator build cache
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: ${{ runner.temp }}/workcell-validator-home/.cache
+          key: validate-buildcache-${{ runner.os }}-${{ hashFiles('go.sum', 'runtime/container/rust/Cargo.lock') }}
+          restore-keys: |
+            validate-buildcache-${{ runner.os }}-
+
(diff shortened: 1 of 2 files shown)


omkhar:
## Closing — empirical savings don't justify the change

Measured cold vs warm CI with the cache working correctly (validator cache hit cleanly: 142MB restored on exact key, `go test ./...` returned all `(cached)`):

| | Validate repository |
|---|---|
| Cold (cache populate) | 11.3m |
| **Warm (cache hit)** | **10.5m** |

That's **~48s (~7%)** — far below the 3–6m the analysis estimated. The compile work the cache accelerates is only a small slice of the 11m validate job; the rest (validator image load, scenario tests, lint, fuzz, codespell/markdownlint) is irreducible and uncacheable this way.

~48s does not justify adding a new external action (`actions/cache`) + cache bind-mount plumbing, and introducing caching into the security-validation pipeline (a posture change). Closing per the repo's "abandon when empirical evidence shows otherwise" rule. The big CI win already landed in #306 (18m → 11.4m).

PR #308, opened 2026-05-30
Release: de-serialize preflight into parallel amd64-repro and smoke jobs

omkhar:
## Summary

Optimization #3 of the CI/deploy series — the release-side counterpart to #306. The release `preflight` job (~30m, the release critical path) ran the amd64 reproducible-build, the container-smoke check, and a disk-reclaim step **serially** inside one job. This hoists the two heavy ones into parallel jobs, mirroring the `preflight-arm64-repro` split from #305.

- **`preflight-amd64-repro`** (new, native ubuntu-latest): reproducibility-verifies the amd64 image, exports its digests as job outputs.
- **`preflight-container-smoke`** (new, ubuntu-latest): runs `container-smoke.sh`.
- `preflight` now `needs` both repro jobs and replaces the inline amd64 build + arm64-merge with a single **assemble** step that builds the combined runtime-preflight manifest from both jobs' verified digests, fail-closed on any missing/malformed digest. `build-arm64-image` and `release` gain `preflight-container-smoke` in `needs` so smoke still gates publication.
- Drop the "Reclaim runner space" prune — it only freed disk for the inline repro build, which now runs on dedicated fresh runners.
- pinned-inputs policy: drop the three obsolete reclaim-step assertions. workflow-lane policy: register the two new release-only lanes (regenerated `workflow-lanes.json`).

## Effect

amd64-repro (~3m) and container-smoke (~4m) overlap the preflight `validate-repo` instead of running after it → **~7m off the preflight critical path**. Same total compute, de-serialized.

## Invariant safety

- The preflight→publish **per-platform digest binding** is preserved: both per-arch digests come from `verify-reproducible-build.sh` (same script, `SOURCE_DATE_EPOCH`, buildkit), and publish still byte-compares the published image against the assembled manifest, fail-closed.
- Reproducible builds stay `--no-cache` + serial double-build.
- New repro/smoke jobs are read-only and ungated; the `release`-environment gating on build/publish is unchanged.

## Testing

`release.yml` only runs on a tag, so the cross-job digest assembly is exercised end-to-end on the **first tagged release after merge** (fail-closed; mirrors the already-proven #305 arm64 split). Local: `actionlint`, `zizmor` (auditor), `check-pinned-inputs.sh`, `verify-workflow-lanes.sh`, `verify-invariants.sh`, `go build/test ./...`, full `pre-merge.sh --profile pr-parity`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -119,11 +119,82 @@ jobs:
           echo "image_manifest_digest=${manifest_digest}" >> "${GITHUB_OUTPUT}"
           echo "config_digest=${config_digest}" >> "${GITHUB_OUTPUT}"
 
+  preflight-amd64-repro:
+    name: Release reproducible image (amd64)
+    # Build and reproducibility-verify the linux/amd64 runtime image in its own
+    # native job (mirroring preflight-arm64-repro) so it runs in parallel with
+    # the rest of preflight instead of serially inside it.  Feeds its verified
+    # image/config digests back via job outputs; preflight assembles the
+    # combined runtime manifest that publish checks against.
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    permissions:
+      contents: read
+    outputs:
+      image_manifest_digest: ${{ steps.repro.outputs.image_manifest_digest }}
+      config_digest: ${{ steps.repro.outputs.config_digest }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - id: source_date
+        name: Compute source date epoch
+        run: echo "epoch=$(git log -1 --pretty=%ct "${GITHUB_SHA}")" >> "${GITHUB_OUTPUT}"
+
+      - uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # v4.0.0
+        with:
+          cache-binary: false
+          driver-opts: image=${{ env.WORKCELL_BUILDKIT_IMAGE }}
+          version: ${{ env.WORKCELL_BUILDX_VERSION }}
+
+      - id: repro
+        name: Verify reproducible amd64 runtime image
+        env:
+          WORKCELL_REPRO_PLATFORMS: linux/amd64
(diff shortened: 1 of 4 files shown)


omkhar:
## Adversarial review — clean

Independent adversarial review (7 dimensions) + re-ran the repo gates (`go test ./internal/metadatautil/...`, `check-pinned-inputs.sh`, `verify-workflow-lanes.sh`, `actionlint`) on the committed tree. **No correctness or security problem found.**

1. **Manifest assembly** — OK. The `jq -n` step produces exactly `.platforms["linux/amd64"|"linux/arm64"].{image_manifest_digest,config_digest}` — the four paths publish reads. Omitting `oci_subject_digest` is safe: grepped all Go/sh/yml — it's consumed only inside the manifest generator + its tests, never read from the preflight manifest downstream.
2. **Digest equivalence** — OK. `preflight-amd64-repro` uses the same `verify-reproducible-build.sh`, `WORKCELL_REPRO_PLATFORMS=linux/amd64`, `SOURCE_DATE_EPOCH` derivation, and buildkit/buildx pins as the old inline step. No drift; mirrors the proven arm64 split.
3. **Gating** — OK. `preflight` needs both repro jobs; `build-arm64-image` and `release` both need `preflight-container-smoke`, so smoke still gates all publication. Fail-closed `!= sha256:*` check before jq. (install-verification no longer transitively needs smoke, but it neither pushes nor publishes — harmless.)
4. **Removed prune** — OK. It only guarded the inline `--no-cache` runtime build, now moved to dedicated fresh runners; no remaining preflight step needs the freed disk.
5. **pinnedinputs.go** — OK. The three removed strings no longer exist in release.yml; the kept `git archive` assertion still matches; nothing else referenced the reclaim step.
6. **Lane policy** — OK. Both new lanes registered (`release-only`/`local_mode:none`), `Release `-name-prefixed so the preflight green-gate won't deadlock; `workflow-lanes.json` consistent.
7. **Other** — OK. New jobs least-privilege (`contents: read`); timeouts present; `--argjson source_date_epoch` gets a valid integer; fail-closed if empty.

**Residual:** the digest binding rests on `preflight-amd64-repro` producing a digest identical to publish's `build-push-action` amd64 build across different fresh runners — the same fail-closed assumption already proven for arm64 (#305) and the prior inline-vs-publish amd64 match, mediated by `rewrite-timestamp` + `SOURCE_DATE_EPOCH`. No defect introduced.

PR #322, opened 2026-06-11
Architectural review iteration 3: fail-closed release gate and verified quickstart

omkhar:
## Summary

Architectural-review iteration 3 (deep adversarial passes on the release/signing pipeline and host-side internals + network egress; both reported **zero high findings**). This PR fixes the two mediums and two lows the release-pipeline pass surfaced:

- **Release gate fail-closed**: the green-main gate passed a tagged commit with zero registered check-runs (tag pushed inside the registration window) and let `stale` conclusions through. Now requires at least one *completed* `github-actions` check suite on `main` for the tagged commit before consulting check-runs (third-party suites like kusari-inspector sit `queued` forever on main and are excluded — verified empirically), and `stale` joins the failing conclusion set.
- **Quickstart actually verifies**: "Option A: verified release bundle" never verified anything; both install options now include the `cosign verify-blob` + `sha256sum -c` steps from the provenance contract.
- **provenance.md attestation toggle**: documented the removed `WORKCELL_ENABLE_GITHUB_ATTESTATIONS` opt-in; now describes the current fail-closed `WORKCELL_RELEASE_NO_ATTEST` opt-out.
- **Zero-byte assets**: `publish-github-release.sh` now rejects empty release assets (`-s` instead of `-f`).

Verified sound in this iteration (no changes needed): all signed artifacts are built in the signing job with fail-closed digest-equality handoffs; keyless cosign identities are effectively exact; egress enforcement is VM-side iptables with the documented IPv6 fail-closed behavior implemented; host publication is argv-only with hooks disabled; every credential read routes through the secretfile gate.

## Validation

- `./scripts/pre-merge.sh --profile pr-parity` green locally
- actionlint + zizmor (check-workflows), shellcheck, codespell, check-pinned-inputs all pass
- Check-suite shape verified against real main commits before changing the gate

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -244,9 +244,31 @@ jobs:
           # tagged commit show failures or unfinished work.  Tag-ancestry
           # alone is not enough: a tag pushed before main CI completes, or
           # against a red main, must not produce a release.
+          # Fail closed when nothing has registered yet: a tag pushed inside
+          # the registration window would otherwise see zero check-runs and
+          # sail through both filters below. GitHub Actions check suites for
+          # main-push workflows register at event delivery, so an empty set
+          # means main CI has not started for this commit. Scope to the
+          # github-actions app on main: third-party app suites (for example
+          # kusari-inspector) can sit queued forever on main commits.
+          suites="$(gh api "repos/${GITHUB_REPOSITORY}/commits/${GITHUB_SHA}/check-suites" \
+            --paginate \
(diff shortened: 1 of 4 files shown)


omkhar:
@codex review

omkhar:
@codex review

omkhar:
Addressed both review findings: switched the quickstart verification to `shasum -a 256 --ignore-missing -c` (matching the install-verification lane — stock macOS lacks GNU sha256sum), and the release-asset preflight now requires `-f && -s` so a non-empty directory fails before any draft assets are deleted.

@codex review

PR #323, opened 2026-06-12
Architectural review iteration 4: delete the dead publish_pr bash block

omkhar:
## Summary

Architectural-review iteration 4: land the dead-code deletion the `scripts/workcell` header promised when `publish_pr_main` became a Go shim, and clean up everything coupled to it.

- Delete the 14 unreachable functions (`publish_pr_usage` through `publish_pr_repo_owned_checks_expected`, including `resolve_existing_file_or_die`/`resolve_existing_executable_or_die`/`run_publish_host_command_in_dir`/`validate_publish_*`, whose only callers were the dead set) — 202 lines.
- Drop the file-level `SC2317,SC2329` shellcheck suppression. The one legitimate cross-file consumer, `session_lookup_root_args` (called from the sourced `sessionctl-shim.sh`), keeps a narrow per-function disable with the reason, matching the `cleanup()` precedent.
- Re-point two `verify-invariants.sh` checks that asserted the *bash shape* of migrated functions at their Go owners (`publishpr.ResolveExistingExecutableOrDie` trusted-path rejection, `publishpr.ValidateBaseName` check-ref-format validation); all other harness targets verified live.
- Regenerate the control-plane manifest (`scripts/workcell` is sha-pinned in it).
- Delete `tests/scenarios/gemini-auth-modes/` — three fixtures nothing has ever referenced; gemini auth-mode selection is covered by the `verify-invariants` harness.

## Validation

- `./scripts/pre-merge.sh --profile pr-parity` green (this deletion exercised every layer: host shellcheck, validator-image shellcheck, control-plane manifest verification, and the invariants harness — each coupling above was caught by a gate and resolved, not suppressed)
- `verify-scenario-coverage` green after the fixture removal
- No dynamic-scope callers: all deleted names grepped across scripts/, tests/, internal/, and the sourced libs


omkhar:
@codex review

omkhar:
Addressed the review finding: the migrated invariant now asserts the `!checkRefFormat(base)` rejection call itself rather than the function signature (and both migrated checks now use the harness's portable `grep -qF --` form — the CI validator's rg fallback shim doesn't support `-qF`).

@codex review

omkhar:
Addressed: the trust-check assertion now includes the leading raw-path negation, anchoring both rejections.

@codex review

omkhar:
Addressed: added `go_function_block_contains_fixed` (the Go analogue of the harness's bash function-block helpers) and scoped both migrated invariants to the `ResolveExistingExecutableOrDie` and `ValidateBaseName` bodies — comment or unrelated-helper text can no longer satisfy them.

@codex review

omkhar:
Addressed: the helper now strips // line comments and single-line block comments from the extracted function body before matching, and the fix was verified against a synthetic function containing only commented-out copies of the needle (rejected).

@codex review

omkhar:
Addressed: the awk filter now tracks open block-comment state across lines; multi-line and trailing block-comment cases verified rejected while real calls still match. (Two of the round's four comments re-reported the line/single-line case fixed in the previous commit.)

@codex review

omkhar:
@codex review

PR #332, opened 2026-06-12
Architectural review iteration 6: close completeness-sweep mediums

omkhar:
## Summary

Iteration 6 of the architectural review: a completeness-critic sweep over every surface the prior five iterations had not examined (rootio, shellproto, cliexit, scenarios, host/launcher, hoststate, colimautil depth, remotevm, scripts/lib, adapter hooks/rules, profiles) found **no high findings** and four mediums — all fixed here:

- **Glued `-c` hook-bypass gap**: the Claude bash-hook guard required whitespace after `-c`, so `git -ccore.hookspath=…` evaded it; regex fixed and the glued form added to the blocked-commands fixture (red→green).
- **Codex nested-CLI rules drift**: `default.rules` forbade the nonexistent `dist/index.js` Gemini path instead of the installed `bundle/gemini.js`, and omitted `provider-wrapper.sh` which managed_config forbids; `default.rules` and `requirements.toml` aligned with the managed config (control-plane parity verifies).
- **Planted synthetic-home**: `trusted-docker-client.sh`'s stable `/tmp/workcell-home-UID` fallback honored pre-planted symlinks/foreign dirs on shared hosts; now refuses anything that isn't an owned real directory (stable path kept for validator build caches).
- **Remote-vm symlink escape** (preview-only, zero production reachability): materialization copied symlinks verbatim, so absolute/`../` targets escaped the contract's host-auditable workspace; escaping targets now fail closed with tests covering rejection and legitimate internal symlinks; manifest written 0600.

All guards touched are documented defense-in-depth or preview paths — the VM+container boundary is unchanged.

## Validation

- `./scripts/pre-merge.sh --profile pr-parity` green locally
- Hook parametric tests (11 blocked / 5 permitted), remotevm tests incl. two new symlink cases, shellcheck, control-plane manifest + parity, pinned inputs all pass

### DIFF
--- a/adapters/claude/hooks/guard-bash.sh
+++ b/adapters/claude/hooks/guard-bash.sh
@@ -291,7 +291,7 @@ mcp_home_control_regex="(^|[[:space:]'\";|&])(~|\$home|/state/agent-home)/\\.mcp
 if grep_command '(^|[^[:alnum:]_./-])(git|/usr/bin/git|/usr/local/libexec/workcell/git|/usr/local/libexec/workcell/core/git|/usr/local/libexec/workcell/real/git)([^[:alnum:]_./-]|$)'; then
   if [[ "${command_lower}" == *"--no-verify"* ]] ||
     grep_command '(^|[[:space:]'"'"'";|&])commit.*[[:space:]'"'"'";|&]-n([[:space:]'"'"'";|&]|$)' ||
-    grep_command '(^|[[:space:]'"'"'";|&])-c[[:space:]]+(core\.hookspath|core\.worktree|include\.path|includeif\.[^=[:space:]]+\.path)=' ||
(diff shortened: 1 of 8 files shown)


omkhar:
@codex review

omkhar:
Addressed the P1: symlink containment now also requires the kernel-resolved target (filepath.EvalSymlinks) to stay inside the resolved workspace root, with the dir/up -> .. chain-escape case added as a test (red before the fix). Dangling links fail closed.

@codex review

omkhar:
Addressed the manifest-permissions finding: chmod 0600 after WriteFile so rewrites over an existing manifest also tighten permissions. (The P1 thread is the pre-fix comment re-anchored by GitHub — the kernel-resolution fix and the dir/up chain-escape test landed in d50b69f.)

@codex review

PR #337, opened 2026-06-13
Simplify: factor duplicated Codex argv scanners into codex_first_subcommand

omkhar:
## Summary

A `/simplify` cleanup of the Codex provider wrapper. The 0.139 + profile-v2
migration left two argv scanners — `codex_managed_profile_applies` and
`codex_app_server_sandbox_args` — that each walked `"$@"` with the **same**
global value-flag skip table to find the first subcommand token. That put the
Codex global-flag list in two places that had to be edited in lockstep on every
Codex release.

## Change (behavior-preserving refactor, net −23 lines)

- Extract one `codex_first_subcommand` helper that resolves the subcommand token
  once into `CODEX_FIRST_SUBCOMMAND` (single home for the global-flag table).
- `codex_managed_profile_applies` collapses to a flat non-runtime deny-list (the
  old runtime allow-arm was redundant with the deny-list + default-inject).
- `codex_app_server_sandbox_args` sets `MANAGED_CODEX_APP_SERVER_ARGS` directly,
  dropping the per-launch `mapfile`/process-substitution fork.
- The codex case calls `codex_first_subcommand` once before both classifiers.

## Validation

- Helper unit matrix confirms behavior is unchanged: profile inject/skip for
  runtime / non-runtime / alias (`a`, `e`, `cloud-tasks`) / `--version` /
  `--help` / default-TUI cases, and app-server `sandbox_mode` per profile.
- Local `pre-merge.sh --profile pr-parity` passed (validate, container-smoke
  incl. the profile-injection + execpolicy paths, reproducible build).

### DIFF
--- a/runtime/container/control-plane-manifest.json
+++ b/runtime/container/control-plane-manifest.json
@@ -232,7 +232,7 @@
       "kind": "runtime-control-plane",
       "repo_path": "runtime/container/provider-wrapper.sh",
       "runtime_path": "/usr/local/libexec/workcell/provider-wrapper.sh",
-      "sha256": "bd5cb8e61aa3925e9d88033aef43e452231f0393e9e92655ae84ca2e2e609bff"
+      "sha256": "d18c45a5833ccba0067cbad251b7822abdfe115dfa9f9da84fb9903863786bc7"
     },
     {
       "kind": "runtime-control-plane",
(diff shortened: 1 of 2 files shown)


omkhar:
@codex review

PR #340, opened 2026-06-14
Simplify: dedup git-root gating + tracked-input error in core.go

omkhar:
## Summary

A `/simplify` cleanup of `internal/metadatautil/core.go` (the control-plane /
build-input manifest generators). Three behavior-preserving changes (net −6
lines):

- `gitTrackedFiles` and `gitTrackedSubset` both opened with the identical
  inside-work-tree + top-level-matches-`rootDir` guard. Extracted
  `gitRootMatches` and call it from both (each keeps its own no-git return shape).
- Inlined the single-use one-line `ensureNoSymlinkPrefix` wrapper at its only
  caller.
- `digestMap` emitted the same long "tracked release input is missing from the
  working tree" message in two branches; hoisted to a `missingTrackedInputMsg`
  const referenced by both.

## Validation

- `go test ./internal/metadatautil` passes unchanged; `gofmt`/`go vet` clean.
- Local `pre-merge.sh --profile pr-parity` passed.

### DIFF
--- a/internal/metadatautil/core.go
+++ b/internal/metadatautil/core.go
@@ -24,6 +24,8 @@ var (
 	hexDigestPattern = regexp.MustCompile(`^[0-9a-f]{64}$`)
 )
 
+const missingTrackedInputMsg = "tracked release input is missing from the working tree; stage the deletion or restore the file before generating a verified build input manifest: %s"
+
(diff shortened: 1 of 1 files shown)


omkhar:
@codex review

PR #341, opened 2026-06-14
Simplify: hoist recompiled regexes to package vars in pinnedinputs.go

omkhar:
## Summary

A `/simplify` cleanup of `internal/metadatautil/pinnedinputs.go`. Unlike the
sibling `provider_bumps.go` (which hoists every regex to a package var), this
file compiled several constant regexes inline on every call. Hoist the worst
offenders:

- `^v\d+\.\d+\.\d+$` (the exact-pinned-release-tag check) was compiled at **five**
  separate call sites → `pinnedReleaseTagPattern`.
- The per-workflow `uses:` scan and the 40-hex commit-SHA check were recompiled
  **inside the per-workflow / per-`uses` loop** (the hottest recompile path) →
  `workflowUsesPattern` and `commitShaPattern`.

Patterns are byte-for-byte identical, so behavior is unchanged.

## Validation

- `go test ./internal/metadatautil` (which validates the Dockerfile/workflow
  pins) passes unchanged; `gofmt`/`go vet` clean.
- Local `pre-merge.sh --profile pr-parity` passed.

### DIFF
--- a/internal/metadatautil/pinnedinputs.go
+++ b/internal/metadatautil/pinnedinputs.go
@@ -19,6 +19,12 @@ import (
 var (
 	workflowPermissionsRE = regexp.MustCompile(`(?m)^permissions:\s+\{\}$`)
 	aptInstallPattern     = regexp.MustCompile(`apt-get install -y --no-install-recommends(?s:(.*?))&&`)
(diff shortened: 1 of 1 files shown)


omkhar:
@codex review

PR #370, opened 2026-07-04
Add reviewed GitHub Actions allowlist (B4 core)

omkhar:
# Summary

Lands the highest-value half of ROADMAP item **B4** (Centralized Tool Pins And
Action Allowlist) per [`docs/improvement-tracks-implementation-plan.md`](docs/improvement-tracks-implementation-plan.md):

- **`policy/allowed-actions.toml`** — a reviewed allowlist of the 12 GitHub
  Actions `owner/repo` identities permitted in workflows.
- **`internal/metadatautil/pinnedinputs.go`** — `check-pinned-inputs` now
  rejects any `uses:` whose `owner/repo` is not on the allowlist, in addition to
  the existing full-commit-SHA requirement. SHA-pinning proves ref integrity;
  the allowlist restricts *which* publishers may run at all, so a new (even
  correctly pinned) action cannot be introduced without a reviewed policy diff.
- The `uses:` scan pattern was extended to also match 3-segment actions
  (`github/codeql-action/analyze`, `anchore/sbom-action/download-syft`), which
  the old pattern skipped — so those are now SHA-checked and allowlist-checked
  too (closing a pre-existing gap). No workflow YAML changed.

## Scope note

This is B4's action-allowlist half. The tool-pin **centralization** (moving the
duplicated actionlint/zizmor/cosign/buildx/... pins into a canonical policy file)
is a larger, separately-reviewable change deferred to **B4b**; the pins are
already cross-checked for drift by `check-pinned-inputs` today.

## Support-claim impact

None. CI supply-chain validation and docs only.

## Validation

- new negative test: an off-allowlist (but SHA-pinned) action is rejected; full
  `CheckPinnedInputs` suite green; `go vet`/`gofmt` clean
- `check-pinned-inputs`, `check-workflows`, control-plane-lockstep invariant,
  `validate-toml`, docs-link, markdownlint, codespell all pass
- `./scripts/pre-merge.sh --profile pr-parity` before publication

### DIFF
--- a/docs/github-workflows.md
+++ b/docs/github-workflows.md
@@ -115,6 +115,14 @@ feeds its digests into the preflight manifest, while the amd64 image is verified
 natively in `preflight`. `check-pinned-inputs` rejects any reintroduction of
 `docker/setup-qemu-action` into the release workflow.
 
+`check-pinned-inputs` also gates the GitHub Actions supply chain: every `uses:`
+must be pinned by full commit SHA, and its `owner/repo` must appear in the
+reviewed allowlist
+[`policy/allowed-actions.toml`](../policy/allowed-actions.toml). SHA-pinning
+proves the integrity of a specific ref; the allowlist restricts which action
+publishers may run at all, so introducing a new action — even a correctly
+pinned one — requires a reviewed change to that policy.
+
 `ci.yml` and `docs.yml` use the same explicit nonroot validator contract when
 they bind-mount the repository: the workflow computes the caller UID/GID,
 passes isolated writable roots, and creates those paths inside the validator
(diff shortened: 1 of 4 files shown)


omkhar:
@codex review

omkhar:
Addressed in 7e2c5456e3: good catch — a quoted step key (`- "uses":`) would evade the raw-line scan. Replaced the line-regex uses extraction with YAML parsing (extractWorkflowUses unmarshals jobs->steps->uses), so quoted keys, dash-less step keys, and odd spacing all resolve to the same `uses` field and cannot bypass the pin/allowlist checks. Added a quoted-key negative test; real workflows still pass.

omkhar:
@codex review

PR #371, opened 2026-07-04
Centralize CI tool pins into reviewed policy (B4b)

omkhar:
# Summary

Completes ROADMAP item **B4** (the pin-centralization half; the action allowlist
half landed in #370) per [`docs/improvement-tracks-implementation-plan.md`](docs/improvement-tracks-implementation-plan.md):

- **`policy/tool-pins.toml`** — one reviewed canonical source for the 7 CI/release
  tool pins (cosign, buildx, buildkit, QEMU, syft, actionlint, zizmor) that were
  otherwise duplicated inline across workflows.
- **`check-pinned-inputs`** now asserts each tool's canonical workflow value
  matches the policy (after the existing cross-file consistency asserts, which
  transitively keep every other workflow copy in lockstep). **syft gains
  consistency coverage it did not have before.** Bumping a tool is now a reviewed
  change to one file.
- **`scripts/update-upstream-pins.sh`** rewrites the policy alongside the
  workflows (before its own `check-pinned-inputs` self-check), so the automated
  bump path stays consistent.

## Support-claim impact

None. CI supply-chain validation, a policy file, and docs only.

## Validation

- new tool-pin policy-drift test; full `internal/metadatautil` suite green
  (existing cross-file mismatch tests preserved — policy block runs after them)
- `check-pinned-inputs` passes on the real repo (policy matches current pins);
  control-plane-lockstep, `validate-toml`, docs-link, `shellcheck`, `codespell`,
  `markdownlint` all pass
- `./scripts/pre-merge.sh --profile pr-parity` before publication

### DIFF
--- a/docs/github-workflows.md
+++ b/docs/github-workflows.md
@@ -123,6 +123,15 @@ proves the integrity of a specific ref; the allowlist restricts which action
 publishers may run at all, so introducing a new action — even a correctly
 pinned one — requires a reviewed change to that policy.
 
+The CI/release tool pins (cosign, buildx, buildkit, QEMU, syft, actionlint,
+zizmor) have a reviewed chokepoint in
+[`policy/tool-pins.toml`](../policy/tool-pins.toml). The values still live inline
+in the workflows; `check-pinned-inputs` asserts every workflow copy equals the
+policy, and `scripts/update-upstream-pins.sh` rewrites the workflows and the
+policy together on a bump. The policy is the human-reviewed record kept in
+lockstep with the workflows, so a pin can never move without a reviewed diff to
+that file.
+
 `ci.yml` and `docs.yml` use the same explicit nonroot validator contract when
 they bind-mount the repository: the workflow computes the caller UID/GID,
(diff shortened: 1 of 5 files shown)


omkhar:
@codex review

omkhar:
Addressed in 25e67e1cb6: good catch — actionlint used requireRegex (first match only), so a second ACTIONLINT_* block could drift. Switched both actionlint version and SHA to requireUniformWorkflowEnv (the same all-copy validation zizmor uses), so every copy in a workflow must be uniform before the policy comparison runs. Full suite + real check-pinned-inputs green.

omkhar:
@codex review

omkhar:
Addressed in 1874a5d276: good point — the tests now derive the mutation from the fixture's actual pin value via a regex on each key line (replacing the value with a sentinel) and t.Fatal if the replacement is a no-op, so a routine upstream-refresh bump won't silently turn a subtest into a no-op or require hand-editing literals.

omkhar:
@codex review

PR #377, opened 2026-07-04
Public contract inventory + drift check (G1 part 1)

omkhar:
# Summary

Implements ROADMAP item **G1 part 1 (Public Contract Inventory)** per
[`docs/improvement-tracks-implementation-plan.md`](docs/improvement-tracks-implementation-plan.md),
extending the existing `operator-contract` drift-check framework (it does **not**
duplicate the CLI-flag/help parity that `validate-operator-contract` already covers):

- **`policy/public-contract.toml`** — the machine-checkable contract: exit codes,
  stable output-line prefixes, `SessionRecord`/`SessionExport` JSON fields, and the
  injection-policy table whitelist. Marked `version = "1"`.
- **`validate-public-contract`** (new `workcell-citools` subcommand →
  `internal/metadatautil.CheckPublicContract`) — a deterministic contract-to-code
  drift check: exit codes ⊆ code, output prefixes ⊆ code (with a shellproto-aware
  match), and **set-equality** for the session-record fields (vs the struct json
  tags) and injection tables (vs the parser's accept guards). Fail-closed.
- **`scripts/check-public-contract.sh`** — the docs-lane wrapper (mirrors
  `check-pinned-inputs.sh`'s `#!/bin/bash -p` + trusted-entrypoint self-sanitizing
  pattern, since it runs `go` on the host), wired into `job-docs.sh`; registered in
  the `HOST_GATE_SCRIPTS` hostile-environment probes, the shellcheck target list, and
  the `docs.yml` path filters (+ regenerated lane manifest).
- **`docs/stability-contract.md`** — extended into the versioned inventory (output
  prefixes, injection tables, session fields, `Contract version: 1`).
- **`man/workcell.1`** — added the missing exit-code **3** (colima-status no-match).

Freeze is deferred to G1 part 2 at v1.0-rc.

## Support-claim impact

None. Adds a CI drift check + documents the existing surface.

## Validation

- 5 Go tests including 4 negative controls (bogus exit code, dropped session field,
  bogus output prefix, bogus injection table — each fails with a naming error)
- `validate-public-contract . policy/public-contract.toml` exits 0 on the real repo
- `go test ./...`, `check-workflows`, `verify-workflow-lanes`, control-plane-lockstep
  (the new policy is doc-referenced), shellcheck, markdownlint, codespell all green
- `./scripts/pre-merge.sh --profile pr-parity` passed (its verify-invariants validates
  the HOST_GATE trusted-entrypoint addition under hostile BASH_ENV / functions / PATH)


omkhar:
@codex review

omkhar:
Addressed in 6e898e6a25, both agreed:
1. (prefix boundary) Replaced the substring match with a key-boundary regex `(?:^|[^A-Za-z0-9_])<prefix>`, so `assurance=` is no longer satisfied by `current_assurance=`/`cache_assurance=`/`autonomy_assurance=` — deleting the real output now fails. Added a substring-only-prefix negative control (`surance=`) proving it.
2. (completeness) Added the documented-and-emitted `host_arch=`, `target_provider=`, and `surviving mutants:` prefixes to both policy/public-contract.toml and the doc table, so they are now enforced. Real-repo check + all 6 tests green.

omkhar:
@codex review

omkhar:
Addressed in 2c31357122: confirmed the self-contamination is entirely in this validator's own doc comments (no other file mentions the prefixes in comments). Excluded internal/metadatautil/public_contract.go from the emitter corpus alongside the existing _test.go exclusion, and locked both with a test. The real-repo check still passes, which proves every prefix is found in a genuine emitter rather than this file's prose.

omkhar:
@codex review

omkhar:
Addressed in 41bc02f4f4, both agreed:
1. (docs trigger) You're right that only docs.yml runs the check. Added cmd/\*\*, internal/\*\*, and scripts/workcell to both docs.yml path blocks (+ regenerated the lane manifest), so removing a SessionRecord tag, changing the injection whitelist, or deleting a stable prefix now starts the docs workflow and runs the drift check.
2. (emitter anchor) Tightened the output-prefix match to require an immediately-preceding quote (`["']<prefix>`), so it only matches a format-string emitter (fmt.Sprintf("record_digest=..."), printf 'record_digest=...') and no longer a bare `local record_digest=` assignment or the `.*record_digest=` sed pattern. Added a direct emitter-anchor unit test proving assignments/sed refs don't satisfy it. Verified all 13 documented prefixes still resolve to real quoted emitters.

omkhar:
@codex review

omkhar:
Two findings, one fixed and one respectfully pushed back:

1. **Exit codes at exit sites (agreed, fixed in 7b1917d040)** — anchored each code to a real exit construct: `os.Exit(N)`, `cliexit.ExitCodeError{Code: N}`, `return N`, shell `exit N`, and the Rust launcher's status literals (`{ 126 }`/`{ 127 }`, `128 + WTERMSIG`, the `= 124` timeout const). A standalone arity/index digit no longer satisfies a documented code; added an exit-anchor unit test (arity/index → not emitted).

2. **Help-text prefixes (already handled — pushing back)** — the quote-anchor added earlier requires the prefix to *immediately* follow a quote, i.e. to *begin* a format-string literal. `internal/authpolicy/usage.go` mentions it mid-sentence (`reports provider_bootstrap_*`, preceded by a space), so it does **not** match — I verified `grep -E '["']provider_bootstrap_' usage.go` returns nothing. The real emitter `printBootstrapSummaryWithPrefix(stdout, summary, "provider_bootstrap_")` in bootstrap.go does match. So removing the emitter while leaving the help text would already fail the check; no non-emitting quoted help/comment source currently satisfies any documented prefix.

omkhar:
@codex review

omkhar:
Addressed in 81ec84794b: agreed — the authoritative gate is allowedRootPolicyKeys (validateAllowedKeys runs before any table parsing), and the later name!= scrape would miss a key dropped there. Reworked checkInjectionTables to set-equal the contract's tables + a new declared scalar_root_keys (version/includes) against the allowedRootPolicyKeys map itself. Added a scalar-root-key negative control (dropping version now fails), and removed the now-unused functionScopedMatches. All 10 tests + the real-repo check pass.

omkhar:
@codex review

omkhar:
Addressed in b8990d398e: confirmed — publish_pr_cli.go's missing := "workspace" plus the shellproto import would have masked a removed emitter. Narrowed the fallback to require a shellproto Field construction (Key: "key") rather than any bare quoted key, and added a unit test proving Field{Key: "publish_pr_url"} satisfies it while a bare missing := "publish_pr_url" does not. Real-repo check + all tests green.

omkhar:
@codex review

omkhar:
Addressed in 76234dbd94, both agreed:
1. (comments) Added stripComments (drops Go/Rust // and /* */ and shell full-line #, keeping shebangs and inline ${#…}) and applied it to both the exit-code and output-prefix corpuses, so an explanatory '// exit 2' comment (hostutil:527) or a prefix mentioned only in prose no longer counts as an emitter. Unit-tested.
2. (table/scalar split) Restored the table-specific assertion: [injection_tables].tables must set-equal the actual documentToInjectionMap 'name != …' + extractCopiesBlocks 'tableName != …' guards, in addition to the tables+scalar_root_keys == allowedRootPolicyKeys union. Moving copies into scalar_root_keys now fails (negative-control test added) even though the flattened union still matches the gate.

omkhar:
@codex review

omkhar:
Addressed in 79ef26165c: agreed — the curated output_lines list left most of the session summary unprotected. Added a [session_show_prefixes] category enforced by set-equality against the 42 key= lines SessionShowLines emits, so renaming, adding, or dropping any session summary line now fails the check (negative control added dropping display_workspace=). This gives the session show --text output the same complete, both-directions protection as the SessionRecord json fields.

omkhar:
@codex review

PR #381, opened 2026-07-04
Add operator egress policy depth: [network] allowlist surface (A1)

omkhar:
# Summary

Implements ROADMAP item **A1 (Egress Policy Depth)** — the delta on the shipped
default-deny per-session egress allowlist. **No change to the shipped enforcement**
(colima `iptables`/`ip6tables` default-deny in `DOCKER-USER`, fail-closed dual-stack).

## What lands

- **Operator `[network]` injection-policy surface** — operators can `extend`
  the per-session allowlist (`allow_endpoints`) and `tighten` it
  (`deny_endpoints`), through the reviewed injection-policy path:
  ```toml
  [network]
  allow_endpoints = ["registry.internal.example:443"]
  deny_endpoints  = ["chatgpt.com:443"]
  ```
  Endpoint grammar mirrors `scripts/colima-egress-allowlist.sh` exactly.
- **No-weakening guarantee** — `renderNetwork` structurally returns only
  endpoint slices; there is no code path from `[network]` to `NETWORK_POLICY`,
  the enforcement mode, or an unrestricted posture. Any mode-shaped key
  (`network_policy`, `mode`, `unrestricted`, …) is rejected as unknown.
  `TestRenderNetworkCannotSetPolicyMode` is the explicit proof.
- **deny wins** — `deny_endpoints` are subtracted last, after every allow
  source (provider/broker/credential/injection), and can only remove.
- **Enforcement-parity label** — `egress_enforcement=allowlist` when the
  colima allowlist actually applies, `=none` otherwise (docker-desktop,
  aws-ec2-ssm, gcp-vm), making the coverage gap explicit.
- **Docs** — new `docs/egress-policy.md` (reviewed artifact + parity table);
  `injection-policy.md` `[network]` key; `invariants.md` mechanism detail;
  `threat-model.md` (extensions flow only through the reviewed path).
- **Contract + tests** — `[network]` added to the public contract;
  fail-closed negative controls (malformed/empty/unknown/non-array, mode-key
  rejection); dry-run scenario asserting the label, deny-wins, and fail-closed.

## Threading note

`[network]` is threaded through `internal/authresolve` and `internal/authpolicy`
(which parse/re-render the policy before injection) so it survives into the
resolved policy; both only pass it through as a table with no mode interpretation.

## Support-claim impact

The shipped default-deny allowlist is unchanged; this adds an operator input
path + labeling, both flowing through the already-certified enforcement helper.
No new enforcement claim; the operator surface cannot weaken the default.

## Validation

- no-weakening test-proven; fail-closed negatives; grammar mirrors the helper
- `go test ./...` green; `validate-public-contract`, lockstep, control-plane
  manifest, shellcheck, dry-run scenario all pass
- `./scripts/pre-merge.sh --profile pr-parity` passed


omkhar:
@codex review

omkhar:
Addressed in f09d677ef4: correct — validate/show only re-rendered the [network] table without applying the endpoint/unknown-key checks, so policy validate could pass a policy launch rejects. Added validatePolicyNetwork wired at every loadPolicyBundle validation entry (root/merged/loaded/policy): it rejects mode-shaped keys (network_policy, mode, …) as unsupported and validates every endpoint. Moved the endpoint grammar into the shared internal/injectionpolicy package so the helper, injection render, and authpolicy validate now use one validator (no triplication). Fail-closed negative controls added (network_policy/mode/malformed/empty/non-array/non-string all rejected). go test ./... + contract + lockstep green.

omkhar:
@codex review

omkhar:
@codex review

omkhar:
Two findings, both valid:

**P1 (deny during bootstrap) — FIXED in 5b44002ff5.** Correct: deny_endpoints only tightened ALLOW_ENDPOINTS, so a denied endpoint stayed reachable while the firewall ran with BOOTSTRAP_ENDPOINTS during a rebuild/first launch — a fail-open in deny-wins. Now INJECTION_DENY_ENDPOINTS is subtracted from BOOTSTRAP_ENDPOINTS too, so deny wins across NEEDS_REBUILD/stale-marker paths (control-plane manifest regenerated).

**P2 (empty allowlist after deny) — agreed, landing as an immediate follow-up.** The fix (empty endpoint list → install a drop-all allowlist) belongs in scripts/colima-egress-allowlist.sh; adding that file pushes this PR past the 25-file shape limit, so per the shape check's own split-concerns guidance I'll land it as a focused follow-up right after this merges. Importantly this is not a security regression today: on an empty list the helper aborts the launch, which is fail-closed (no session, no egress) — the follow-up just turns that abort into a clean drop-all so a tighten-to-nothing policy launches with zero egress instead of erroring. I've implemented + locally verified the drop-all render (DROP + established/related, no ACCEPTs) and will open it immediately.

omkhar:
@codex review

omkhar:
Addressed in 77e783ec1a: agreed — contracting egress_enforcement= as stable while leaving stability-contract.md at prev_digest= is drift, and updating that doc would exceed the 25-file shape limit. Removed egress_enforcement= from public-contract.toml [output_lines] so nothing is designated stable-but-undocumented; the label is still emitted and documented in egress-policy.md. The immediate egress follow-up (with the empty→drop-all helper change) adds egress_enforcement= to both public-contract.toml and stability-contract.md together, so the stable designation and its documentation land in lockstep.

omkhar:
@codex review

omkhar:
Addressed in 08e7dad. Agreed the low-level 'Missing endpoint list' crash was wrong for the tightest policy. I resolved it fail-closed rather than as a drop-all launch: added fail_empty_egress_after_deny guards after both the session and bootstrap deny subtractions (colima+strict only) that abort with an actionable diagnostic naming the cause (deny removed every endpoint) and the remedy (drop a deny or widen allow_endpoints). Reasoning: a session with zero allowed egress cannot reach any provider API or registry — it would launch non-functional — so failing closed with clear guidance is the correct default, and it stays strictly fail-closed (no session, no egress installed). A deliberate offline/drop-all session mode could be a future opt-in, but silently launching a broken session isn't the right default. The subtract logic itself is unit-tested (SubtractEndpointList), and the empty→abort is a trivial guard on its output; the existing assurance dry-run scenario passes with the guard in place.

omkhar:
@codex review

omkhar:
Addressed in e10c183a8b: correct — the bootstrap empty-after-deny guard ran at the subtraction site before NEEDS_REBUILD was known, so a policy denying all bootstrap hosts while relying on an already-prepared image failed even though bootstrap egress is never applied. Moved the empty check to the two bootstrap-apply sites (both already gated on colima+allowlist and only reached when a rebuild/prepare actually applies bootstrap egress), so deny-all-bootstrap launches cleanly on a prepared image and still fails closed when a rebuild needs the emptied list. The session-allowlist guard stays unconditional (session egress always applies). shellcheck + assurance dry-run scenario + manifest all green.

omkhar:
@codex review

omkhar:
Addressed in 444664e440: correct — DNS hostnames (and IPv6 literal hex) are case-insensitive, but SubtractEndpointList compared raw tokens, so CHATGPT.COM:443 in deny_endpoints failed to remove the provider's chatgpt.com:443, leaving a denied endpoint reachable. Now the comparison key is lower-cased on both sides so deny wins regardless of casing (kept endpoints keep their original form; the helper resolves case-insensitively anyway). Added case-insensitive test cases to TestSubtractEndpointListRemovesDeniedEndpoints.

omkhar:
@codex review

omkhar:
Addressed in 46f69b36cc: correct — chatgpt.com:0443 validates as port 443 but the string compare missed it. Rather than patch per surface form, I replaced the comparison with a canonicalEndpoint key via net.SplitHostPort that lower-cases the host, renders IP literals canonically (so compressed IPv6 matches too), and strips leading zeros from the port. That closes the deny-wins canonicalization bypass class at the root (case + port + IPv6 form). Added leading-zero-port and compressed-IPv6 test cases.

omkhar:
@codex review

omkhar:
Addressed in 1bf741aeea: correct and important — the resolver whitelisted [network] but didn't validate it, so a resolver-backed credential plus a mode-shaped/malformed [network] could stage credentials and write a resolved policy before the later render rejected it. Added validateNetworkPolicy in authresolve, called right after the policy loads and before any credential resolution, using the shared allow_endpoints/deny_endpoints grammar (net effect: the resolver now fails closed on the same inputs authpolicy/renderNetwork reject, ahead of any staging). The logic mirrors the tested authpolicy.validatePolicyNetwork and the shared injectionpolicy.ValidateEgressEndpoint; I verified it rejects network_policy and malformed endpoints while accepting valid allow/deny lists.

omkhar:
@codex review

omkhar:
Addressed in e2fcb5e25d: correct — chatgpt.com. is the same DNS host as chatgpt.com. Strip trailing dots from the host in canonicalEndpoint. Between this, the earlier host lower-casing, IP-literal canonicalization, and port normalization, the canonical key now exhausts the host spellings the grammar permits: ValidateEgressEndpoint restricts hosts to ^[A-Za-z0-9.-]+$ and already rejects a leading dot and any '..', so case, trailing-dot, IPv6 form, and leading-zero ports are the complete set of semantically-equal variants — deny-wins now holds across all of them. Trailing-dot test case added.

omkhar:
@codex review

omkhar:
Addressed in 7dd912a026: correct — the bracketed-host check only verified the char-set, so [1.2.3.4]:443 or [::::]:443 passed validation but would fail at ip6tables -d mid-launch. Now the bracketed body is parsed with net.ParseIP and required to be a real IPv6 (nil and IPv4-in-brackets rejected), so these fail closed at workcell policy validate/render. Since every operator endpoint passes this validator before reaching the helper, a malformed bracketed literal can no longer reach ip6tables. Added [1.2.3.4]/[::::]/[nothex!] rejection cases. With this, the endpoint validation + canonicalization surface (host case, trailing dot, real IPv6 literal, IPv6 form, port normalization) is comprehensively fail-closed.

omkhar:
@codex review

omkhar:
Addressed in e03dbc10fe: correct — renderNetwork ran after renderDocuments/renderCopies/renderSSH had already written into outputRoot, so a malformed [network] on a policy that also declares documents/copies left the staged bundle behind. Moved the side-effect-free renderNetwork validation to immediately after the policy loads, before any filesystem-writing render step — mirroring the resolver fix, so both the resolver and the bundle-render paths now fail closed on an invalid [network] before staging anything.

omkhar:
@codex review

omkhar:
Addressed in 585f44e721: correct — a dotted-numeric host like 999.999.999.999 passed the label check but the helper routes N.N.N.N to iptables -d as an address. Now a dotted-numeric host must parse as a real IPv4 via net.ParseIP, completing IP-literal parity with the bracketed-IPv6 check. The endpoint validation surface is now comprehensively fail-closed: port range, host-label grammar (no leading/double dots), real bracketed IPv6, and real dotted IPv4; plus the canonical comparison (case, trailing dot, IPv6 form, port normalization) for deny-wins. This PR has now run 13 review rounds; the security core (no-weakening, validate-before-staging in both the resolver and bundle paths, deny-during-bootstrap, canonicalized deny-wins, full endpoint validation) is complete and every finding has been fixed.

omkhar:
@codex review

omkhar:
Pushed a9c5966: a mechanical trim to fit the 1200-line PR-shape budget — condensed the verbose per-fix comments and removed redundant endpoint test cases now covered by the shared-validator grammar test (one representative case per class kept). No behavior or coverage change; all Go tests, the assurance dry-run scenario, contract, and lockstep stay green.

omkhar:
@codex review

omkhar:
Pushed cc84177: condensed the egress docs to bring the diff under the 1200-line PR-shape limit (injection-policy.md now points to egress-policy.md rather than duplicating the grammar/parity detail). Added the approved-heavy-ci label so the reproducible-build and other heavy lanes run for this security-critical change instead of skipping. No behavior change.

omkhar:
@codex review

omkhar:
@codex review

omkhar:
Addressed in 196a20bf2c: correct — adding network to the contract's accepted tables made docs/stability-contract.md stale (it still said exactly four tables). Updated it to five tables including network, noting it only carries allow_endpoints/deny_endpoints and cannot change the network-policy mode. To hold the 25-file PR-shape budget I deferred the threat-model.md [network] mitigation bullet (fully covered by egress-policy.md and invariants.md) to the immediate egress follow-up PR. Contract drift-check, doc-links, and markdownlint stay green.

omkhar:
@codex review

omkhar:
@codex review

omkhar:
Addressed in aa225d72e5: correct and important. deny_endpoints subtracts at the host:port endpoint-entry level, but enforcement resolves the remaining allowlist to IPs and matches on IP — so a denied hostname sharing an IP with an allowed endpoint (two names behind one CDN) can still be reachable at the IP layer. I documented this precisely in egress-policy.md: deny drops the endpoint entry, not the address, and fully blocking a host also requires dropping allow_endpoints that resolve to it. True IP-resolved deny with overlap-fail (resolve denies to IPs and abort/subtract on overlap) is a larger enforcement change — deferred to the egress follow-up rather than bundled here, since it changes the enforcement path and has surprising side effects (denying one CDN name would break others on the same IP).

omkhar:
@codex review

omkhar:
@codex review

omkhar:
@codex review

omkhar:
Addressed in ac521f795b:

P2 (host-side bootstrap deny): correct — deny_endpoints subtracts host:port entries and enforcement is IP-level over VM/container egress, so it tightens the bootstrap build container's egress but does NOT gate the launcher's own host-side release-URL curls during an image rebuild. Rather than piecemeal caveats, I added one explicit scope note in egress-policy.md covering both this and the earlier IP-sharing boundary, and corrected the scripts/workcell bootstrap-subtraction comment to say it tightens the build-container VM egress, not host-side fetches. A true fail-closed host-side gate (resolve denies to IPs / abort the rebuild when a required release host is denied) is queued for the egress follow-up.

P3 (contract): added egress_enforcement= to policy/public-contract.toml [output_lines].prefixes and the docs/stability-contract.md output-line table so a rename/removal is caught by the parity check; regenerated the control-plane manifest. Contract drift-check, lockstep, and manifest verification are green.

omkhar:
@codex review

omkhar:
Addressed in ff9bd05559: correct — the empty-after-deny guard I added treated an emptied session allowlist as fatal even under --prepare-only, which prewarms the image and exits without a session, so prepare-only + a deny that removes the provider endpoints aborted before the image could be prepared. Gated the fatal check on PREPARE_ONLY -ne 1 so it fires only for an actual session launch (colima + allowlist); prepare-only now prewarms and exits regardless of the emptied session allowlist. bash -n, shellcheck, control-plane manifest, and lockstep are green.

omkhar:
@codex review

omkhar:
Addressed in 7dccd28e61: right — gating the fatal guard wasn't enough; restore_runtime_egress still ran the helper apply with an empty ALLOW_ENDPOINTS before the prepare-only exit, and the helper rejects an empty list. I guarded restore_runtime_egress to skip the apply when ALLOW_ENDPOINTS is empty on the allowlist path, leaving the tight build-time egress in place (I did NOT use clear, which means unrestricted — that would weaken the posture). An actual launch never reaches this empty because it aborts at the earlier empty-egress guard, so the empty case is prepare-only only. bash -n, shellcheck, manifest, lockstep, contract, and mdlint are green. The empty→drop-all helper mode that would let prepare-only leave an explicit deny-all remains queued for the egress follow-up.

omkhar:
@codex review

omkhar:
@codex review

omkhar:
@codex review

omkhar:
@codex review

omkhar:
@codex review

omkhar:
Pushed 293c1ed: fixes TestCheckPublicContractRejectsBogusOutputPrefix, whose mutation anchor was the last output_lines prefix — adding egress_enforcement= moved it, so I retargeted the anchor. go test ./... is fully green now (I had omitted internal/metadatautil locally). No behavior change.

omkhar:
@codex review

omkhar:
Addressed in 9bd4478: correct — the standalone docs/egress-policy.md operator artifact wasn't cited in operator-contract.toml or requirements.toml, so the contract inventory and release review wouldn't track the egress surface. Rather than register a new doc (which would exceed the 25-file PR-shape limit), I folded the egress mechanism, [network] surface, no-weakening invariant, scope, and enforcement-parity table into docs/injection-policy.md — which is already inventoried in operator-contract.toml and requirements.toml FR-002 — and deleted the standalone file. The [network] evidence (test-assurance-dry-run.sh) is already in requirements.toml FR-001. Repointed invariants.md and stability-contract.md to the injection-policy section. So the egress surface is now tracked via injection-policy.md's existing registration. go test ./..., doc-links, mdlint, contract, lockstep, and manifest are all green.

omkhar:
@codex review

PR #382, opened 2026-07-05
^F Treat an empty egress allowlist as deny-all (A1 follow-up)

omkhar:
Follow-up to #381 (A1 Egress Policy Depth), closing two deferred items.

## Empty allowlist → deny-all (drop-all)
`scripts/colima-egress-allowlist.sh` `plan`/`apply` rejected an empty endpoint
list with "Missing endpoint list". Now only a **missing argument** is a usage
error; an **explicit empty list** is a valid deny-all — `build_allowlist_rules`
already emits just the ESTABLISHED/RELATED accept and the final `DROP` for it.

The `--prepare-only` restore path now applies this, so a prewarmed idle VM is
left with **no egress** rather than the broader build-time allowlist (`clear`
would mean *unrestricted*, so it is not used). An actual session with an emptied
allowlist still aborts early at the empty-egress guard with its diagnostic.

Added a dry-run scenario asserting the empty plan emits `DROP` and no
per-endpoint `ACCEPT`, with a non-empty negative control. Control-plane manifest
regenerated.

## Threat-model control
Recorded the default-deny per-session egress allowlist in threat-model.md's
Controls list (it mitigates the listed "uncontrolled egress" abuse path),
completing the note deferred from #381 for the PR-shape budget.

## Validation
`go test ./...`, shellcheck (both scripts + scenario), control-plane manifest,
lockstep, doc-links, and markdownlint all green locally.

Remaining A1 deferrals (own follow-up): IP-resolved deny with overlap-fail, and
the host-side bootstrap fail-closed gate — both documented as scope boundaries
in injection-policy.md.

### DIFF
--- a/docs/threat-model.md
+++ b/docs/threat-model.md
@@ -49,6 +49,11 @@ published provenance story.
 ## Controls
 
 - dedicated VM plus container boundary
+- default-deny per-session egress allowlist on the managed colima path
+  (fail-closed, dual-stack `iptables`/`ip6tables` in `DOCKER-USER`); operator
+  extensions flow only through the reviewed injection-policy `[network]` surface,
+  which can tighten or extend the allowlist but never disable it or change the
+  network-policy mode (see [injection-policy.md](injection-policy.md#network-egress-network))
 - explicit runtime profiles with separate assurance labels
 - masking and re-seeding of provider control-plane files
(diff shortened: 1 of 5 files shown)


omkhar:
@codex review

PR #383, opened 2026-07-05
^F Add Rust cargo-fuzz targets for exec-guard classifiers (A3)

omkhar:
Implements the Rust half of **A3: Fuzzing Expansion** (A3-Go already landed).

## Targets (three named surfaces from the roadmap)
- **path_classification** → `classify_protected_runtime_path`, `path_points_to_dynamic_loader`, `classify_loader_target`
- **env_filtering** → `path_from_env_entries`, `env_has_unsafe_git_override`, `resolve_command_via_path_value`
- **git_config_parsing** → `git_config_spec_is_blocked`, `git_config_key_is_blocked`, `git_config_spec_value_is_explicit_safe`

Each has a checked-in seed corpus (`fuzz/corpus/<target>/`, real paths / env entries / git-config specs).

## Exposure (security-conscious)
Private classifiers are exposed to the fuzzer ONLY via `#[cfg(fuzzing)] pub mod fuzz_api` (thin wrappers that discard the private `ProtectedRuntime` type). cargo-fuzz sets `--cfg fuzzing`, so this surface does not exist in the shipped cdylib, `cargo test`, or the normal build. `crate-type` gains `rlib` (cdylib unchanged).

## CI lane
`.github/workflows/fuzz.yml` gains a `rust-fuzz` job (schedule + workflow_dispatch): installs nightly + cargo-fuzz, runs each target for a bounded `-max_total_time`, fails on crash, uploads artifacts. Because the exec-guard crate pins `vendored-sources` for reproducible release builds and `libfuzzer-sys` is not vendored, the lane does an ephemeral `crates-io-remote` source swap for the non-shipping fuzz build only (mirroring the Go lane's `go mod download`) — the committed vendored config is untouched. `docs/fuzzing.md` documents local runs + crash triage/minimization.

## Validation
`cargo test` (17+12 pass — `fuzz_api` does not leak), `cargo clippy --all-targets` (undocumented_unsafe_blocks clean), `cargo fmt --check`, normal `cargo build` all green locally. The target bodies were replayed against every seed + adversarial inputs (empty/all-NUL/invalid-UTF8/pathological) with 0 panics.

**The libFuzzer-linked `cargo fuzz build` + smoke-run runs in CI, not locally**: crates.io is intentionally sandboxed here (vendored-sources), so the CI `rust-fuzz` lane is the build source of truth — triggered via workflow_dispatch on this branch to verify.


omkhar:
@codex review

omkhar:
@codex review

omkhar:
Update: the scheduled fuzz lane's env_filtering target found a real panic in the shipped exec-guard — git_config_key_is_blocked / git_config_key_has_prefix_and_suffix sliced keys by byte offset (key[..6], key[..prefix.len()], key[key.len()-suffix.len()..]), which panics when a multibyte UTF-8 char in a git-config key straddles that index (reachable via GIT_CONFIG_PARAMETERS / -c). Fixed in da2d971 with char-boundary-safe str::get, plus a regression unit test and a corpus seed. Also fixed the lane's crates.io source swap (it pointed at a self-referential crates-io source). cargo test 18+12 / clippy / fmt green; re-running the lane to confirm.

omkhar:
@codex review

omkhar:
Addressed the two reproducibility findings: fixed the local override doc (remove the vendored replace-with, don't add a duplicate crates.io source), pinned libfuzzer-sys ==0.4.13, committed the CI-resolved fuzz/Cargo.lock, and enforced it with a cargo metadata --locked drift-check (cargo fuzz build has no --locked flag). The scheduled fuzz lane is green: real cargo fuzz build + all three targets run on their corpora, and the lockfile drift-check passes. The env_filtering target also found and I fixed a real exec-guard panic (byte-slicing git-config keys on non-char boundaries) with a regression test + corpus seed.

omkhar:
@codex review

omkhar:
Pinned the fuzz nightly to a dated toolchain (WORKCELL_RUST_FUZZ_NIGHTLY=nightly-2026-07-02, the verified-green nightly) across the rustup install, metadata --locked, and fuzz run, and updated docs/fuzzing.md to match — so the lane no longer tracks the moving nightly alias. Together with the exact libfuzzer-sys pin + committed Cargo.lock + metadata --locked drift-check, the Rust fuzz lane is now reproducible. Re-running to confirm the pinned toolchain installs green.

omkhar:
@codex review

omkhar:
Pushing back on the P1 (unsigned commit): this is a false positive. `a6da966684a379e85653f0f9ca5a195126474891` is not one of this PR's commits — it is GitHub's ephemeral test-merge commit, which GitHub auto-generates unsigned and never merges (the current test-merge is 94bfa057, regenerated on every push; a6da9666 is a stale one). Every actual commit on this branch is GPG-signed with key DA5A8E9F536C42FD and shows verified=true/valid via the GitHub API:

- 17d10a9778 verified=true (valid)
- 1af34e195e verified=true (valid)
- 8086530de6 verified=true (valid)
- 9a864f808b verified=true (valid)
- f6cf5e5dd2 verified=true (valid)
- da2d9714ac verified=true (valid)
- 644a2e7ccd verified=true (valid)
- b703c10bf3 verified=true (valid)
- e6fc640e7b verified=true (valid)
- 85bd5d1d3c verified=true (valid)
- 9f658d8246 verified=true (valid)

Unverified commit count on the PR: 0. The merge itself will produce a fresh GitHub-signed merge commit (verified), as it did for #381/#382. There is nothing to re-sign.

omkhar:
@codex review

omkhar:
@codex review

omkhar:
@codex review

PR #384, opened 2026-07-05
^F Add exec-guard syscall-shim performance baselines (C5)

omkhar:
Implements **C5: Syscall-Shim Performance Baselines**.

## What
A dependency-free C microbenchmark (`scripts/bench/exec-guard-bench.c`) + shell driver (`scripts/bench/run-exec-guard-bench.sh`) that time the exec-guard's **allow path** for the interposed `execve`/`execv`/`execvp`/`posix_spawn` family against a benign target (`/bin/true`), `LD_PRELOAD`-ing the built cdylib vs the unhooked libc baseline. The delta is the guard's per-call classification overhead. Any failed launch aborts with no numbers, so a blocked exec can never masquerade as a measurement.

## Optional lane
`.github/workflows/bench.yml` (workflow_dispatch + weekly schedule, non-PR-blocking) builds the cdylib offline from vendored sources, runs the driver with two full passes (the stability gate), and uploads `exec-guard-bench-results`. Registered in `policy/workflow-lane-policy.json` / `policy/workflow-lanes.json` / `policy/retention-policy.json` (+ docs mirror).

## Numbers
The LD_PRELOAD interposition is **Linux-only**, so the reviewed hooked-vs-unhooked numbers come from the `bench.yml` lane, not a local run. `docs/syscall-shim-benchmarks.md` currently carries **explicit placeholders** ("not measured numbers, do not cite until replaced"); I will dispatch the lane and transcribe the two-run results before this merges.

## Validation
Harness compiles+runs the baseline on darwin (verified); shellcheck/shfmt/bash -n, `check-workflows.sh` (actionlint+zizmor+lane+retention), doc-links, markdownlint, and the Rust crate's `cargo test`/clippy/fmt all green locally.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,0 +1,71 @@
+name: Bench
+
+# Optional, non-PR-blocking microbenchmark lane for the workcell_exec_guard
+# LD_PRELOAD shim (runtime/container/rust). It builds the exec-guard cdylib,
+# then times the guard's ALLOW-path classification overhead on the interposed
+# exec/spawn family calls against the unhooked libc baseline
+# (scripts/bench/run-exec-guard-bench.sh), repeating the measurement across two
+# runs to demonstrate the numbers are stable. Results are printed to the job log
+# and uploaded as an artifact; the reviewed baseline numbers and methodology
+# live in docs/syscall-shim-benchmarks.md. This lane is scheduled and on-demand
+# only -- it never runs on the PR path.
+
+on:
+  schedule:
+    - cron: "0 8 * * 1"
+  workflow_dispatch:
+
+permissions: {}
+
+defaults:
+  run:
+    shell: bash --noprofile --norc -euo pipefail {0}
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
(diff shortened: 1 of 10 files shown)


omkhar:
@codex review

omkhar:
Updated: the bench.yml lane ran on Linux and I transcribed the measured baselines into docs/syscall-shim-benchmarks.md — exec-guard allow-path overhead is ~440us/+76% per exec* and ~205us/+42% per posix_spawn, with cross-run hooked-median spread <=0.3% (the two-run stability gate). Reverted the temporary pull_request trigger, so the lane is schedule + workflow_dispatch only (optional, non-PR-blocking) as designed.

omkhar:
@codex review

omkhar:
Both methodology P2s fixed and re-measured. The corrected allow-path classification overhead is ~265us/+48% per exec* and ~57us/+12% per posix_spawn (the earlier figures were higher because they wrongly folded in the per-child .so-load cost the LD_PRELOAD scrub now removes); execvp now exercises PATH search against a controlled PATH. Cross-run hooked-median spread <=1.0% — stability gate met. Temp pull_request trigger reverted; lane is schedule + workflow_dispatch only.

omkhar:
@codex review

PR #388, opened 2026-07-05
^f Support-bundle redaction core (G2, 1/3)

omkhar:
**G2: Support Bundle Command — part 1 of 3** (split for PR-shape limits). This lands the self-contained, tested redaction core first; the collector (part 2) and the `workcell support-bundle` CLI + docs (part 3) follow on top.

## What
`internal/supportbundle/redact.go` — the redaction layer for the forthcoming support bundle. Masks secrets (JWTs, GitHub/OpenAI/Google/AWS/Slack tokens, PEM private keys, `Bearer` headers, secret-named `key=value`) and rewrites the operator home prefix to `~` so no username or credential material can reach a bundle.

## Tests
`redact_test.go` proves each pattern is masked plus an over-redaction guard (sha digests / enum values survive). Test fixtures assemble secret shapes from fragments at runtime so no literal token appears in source (GitHub push protection) while still exercising every pattern.

Leaf library, no caller yet — builds + `go test ./internal/supportbundle/` green, `go vet` clean, gofmt clean.

### DIFF
--- a/internal/supportbundle/doc.go
+++ b/internal/supportbundle/doc.go
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+// Package supportbundle implements `workcell support-bundle`, the host-side
+// diagnostics collector (roadmap item G2).
+//
+// It gathers six evidence classes needed to diagnose install, policy, target,
+// provider, and runtime failures:
+//
+//   - install:        launcher presence/path, repo root, host os/arch, version
+//   - policy:         repo policy inventory + user injection-policy presence
+//   - target:         state-root layout, colima profile directories/markers
+//   - providers:      per-provider adapter presence + credential container path
+//   - sessions:       durable session-record summaries (never workspace bodies)
+//   - audit_pointers: audit-log path metadata (path/size/mtime; never content)
+//
+// The bundle is a single deterministic JSON document: struct field order is
+// stable, every slice is sorted, and the only time-varying field
+// (generated_at) is injected through Config so golden tests are hermetic.
+//
+// Redaction is the security core (see redact.go). The collector NEVER reads
+// credential-file contents, workspace files, or log bodies — it records paths,
+// presence, and metadata only. As defense-in-depth every collected string is
+// additionally passed through the secret redactor before it enters the bundle.
+package supportbundle
--- a/internal/supportbundle/redact.go
+++ b/internal/supportbundle/redact.go
@@ -56,6 +56,9 @@ var (
 	awsKeyRe         = regexp.MustCompile(`(?:AKIA|ASIA)[0-9A-Z]{16}`)
 	openaiKeyRe      = regexp.MustCompile(`sk-[A-Za-z0-9_-]{20,}`)
 	googleKeyRe      = regexp.MustCompile(`AIza[0-9A-Za-z_-]{20,}`)
+	// Google OAuth access tokens (ya29.) from Gemini / gcloud ADC flows; the repo
+	// classifies these as credentials in metadatautil.credentialPatterns.
+	googleOAuthRe = regexp.MustCompile(`ya29\.[0-9A-Za-z_-]+`)
 )
 
 type secretReplacer struct {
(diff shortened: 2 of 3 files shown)


omkhar:
@codex review

omkhar:
Fixed the P1: the key=value value group now matches double-quoted (space-tolerant), single-quoted, or unquoted values and redacts the whole value, so password='hunter2' and password="two words here" are fully masked. Added regression cases for both forms (asserting no fragment like "words here" survives).

omkhar:
@codex review

omkhar:
Fixed both: (P1) the key-name match now allows the secret-bearing stems (password/passwd/passphrase/secret/token/credential) anywhere in the identifier, so secret_key=, SECRET_KEY_BASE=, app_password_field:, refresh_token= are masked — with a new negative test asserting auth_status/public_key/cache_key/sort_key stay legible; (P2) quoted values now honor backslash escapes so password="abc\"def" is masked whole. All redaction tests + the over-redaction guard pass.

omkhar:
@codex review

omkhar:
Fixed both: (P1) pemPrivateKeyRe now runs before the key=value rule, so a PEM block as a secret-named value (private_key=-----BEGIN...) is redacted whole instead of being truncated at the first space; (P2) slackTokenRe now also matches xapp- app-level tokens. Added regression tests for the PEM-in-assignment and xapp- cases; all redaction + over-redaction-guard tests pass.

omkhar:
@codex review

omkhar:
Fixed both: (P2a) the key=value rule now allows an optional backslash before the key's closing quote and matches an escaped-double-quote value, so password=\"x\" and {\"password\":\"x\"} are masked whole; (P2b) the Slack matcher now covers xwfp- workflow tokens and xoxe. rotation tokens in addition to xox[baprs]-/xapp-. Regression cases added for all forms; tests pass.

omkhar:
@codex review

omkhar:
Fixed: added a ya29. Google OAuth token pattern (mirroring metadatautil.credentialPatterns) so Gemini/gcloud ADC access tokens are masked, with a regression case.

omkhar:
@codex review

omkhar:
Fixed: the home rewrite now matches only at a path-component boundary (separator, end-of-string, or a non-path delimiter), so Home=/Users/al no longer corrupts /Users/alice into ~ice; added regression cases including the prefix-collision, whitespace/quote boundaries, and the .-continues-a-component case.

omkhar:
@codex review

omkhar:
Fixed both: single-quoted mapping keys ({'password':'x'}) now match, and the home-boundary set now includes backslash so an exact home before an escaped quote is rewritten. Regression cases added.

I also documented the layering explicitly on `Redactor.String`: this pattern masking is **defense-in-depth**, and the support bundle's **primary** guarantee is **collect-by-construction** — the collector never reads credential-file contents, workspace files, or log bodies, only structured metadata, so secrets don't enter the bundle to begin with. A regex over arbitrarily-escaped free-form text cannot be exhaustive by construction; any field that could carry an adversarially-encoded secret is kept out at the collector rather than relied on this backstop. I've covered the common secret formats and quoting/escaping forms (10 fixes across this review); further exotic-escaping variants of the free-form backstop are dispositioned against that documented boundary.

omkhar:
@codex review

omkhar:
Dispositioning this P2 (not fixing) — recording the decision per the convergence policy:

Codex has issued an LGTM at this head (`Reviewed commit: 59c2774e5c`, "Didn't find any major issues"). This P2 is the 11th free-form quoting/escaping variant surfaced in this review (Bash `%q` `\'...\'`). It falls squarely under the **defense-in-depth boundary I documented on `Redactor.String` in this same PR**: the support bundle's primary guarantee is **collect-by-construction** — the collector never reads credential-file contents, workspace files, or log bodies, only structured metadata, so a Bash-`%q`-escaped password assignment is not a value the bundle collects. The pattern redactor is an explicitly non-exhaustive backstop for the few free-form strings (paths, error text); a regex over arbitrarily shell/JSON-escaped text cannot be made complete, and chasing each new escaping form would restart the review loop indefinitely.

The 10 fixes already landed cover every token family and the common quoting/escaping forms. Per the loop policy (LGTM at head; recent rounds are non-blocking P2s with no P0/P1), I'm merging on the documented boundary rather than iterating further on exotic free-form escaping. If a concrete case shows such a value actually reaching a collected field, that's a collector fix (keep it out), not a redactor-regex fix.

PR #389, opened 2026-07-05
^F Support-bundle collector + leak-proof + golden (G2, 2/3)

omkhar:
**G2: Support Bundle Command — part 2 of 3.** Builds on the merged redaction core (PR-1); the `workcell support-bundle` CLI + docs follow in part 3.

## What
`internal/supportbundle/{collect,bundle}.go` — the host-side collector. It gathers the six evidence classes (install, policy, target, providers, sessions, audit-pointers) as **structured metadata only** — never reading credential-file contents, workspace files, or log bodies (collect-by-construction) — and runs every field through the merged redaction backstop. Deterministic single-JSON shape.

## Tests (the security exit-gate)
- `leak_test.go` — seeds known secrets into **real collected fields** (session workspace path, audit-log body, injection-policy body) and asserts none survive the full collect + marshal pipeline.
- `bundle_test.go` — `TestGoldenBundleShape` pins the deterministic shape (temp home → stable `~` paths, fixed clock); `TestBundleDeterministic` asserts byte-stability.
- `collect_test.go` — one scenario per failure class (install/policy/target/provider/runtime).

`go test ./internal/supportbundle/` green (incl. golden + leak-proof), `go vet` clean. Token fixtures are assembled from fragments so no literal secret appears in source (push protection). 1180 lines, 0 binaries.

### DIFF
--- a/internal/supportbundle/bundle.go
+++ b/internal/supportbundle/bundle.go
@@ -1,0 +1,122 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+package supportbundle
+
+import (
+	"encoding/json"
+	"runtime"
+	"time"
+)
+
+// SchemaVersion versions the bundle JSON shape. Bump it on any field
+// addition/rename/removal so consumers and the golden test can gate on it.
+const SchemaVersion = "1"
+
+// toolName is the stable tool identifier embedded in the bundle.
+const toolName = "workcell support-bundle"
+
+// Config carries every host input the collector needs. Injecting these (rather
+// than reading process env directly) keeps collection hermetic and golden tests
+// reproducible.
+type Config struct {
+	// RepoRoot is the workcell checkout / install root (ROOT_DIR).
+	RepoRoot string
+	// LauncherPath is the absolute path to the scripts/workcell launcher.
+	LauncherPath string
+	// RealHome is the operator home directory used for redaction and to
+	// locate the user injection policy.
+	RealHome string
+	// WorkcellStateRoot and ColimaStateRoot are the two durable state roots
+	// (WORKCELL_STATE_ROOT, COLIMA_STATE_ROOT).
+	WorkcellStateRoot string
+	ColimaStateRoot   string
+	// HostOS and HostArch describe the host; empty falls back to the build's
+	// runtime.GOOS / runtime.GOARCH.
+	HostOS   string
+	HostArch string
+	// Now is the bundle timestamp. A zero value is treated as unset and the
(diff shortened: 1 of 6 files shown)


omkhar:
@codex review

omkhar:
Fixed: the session summary and target-profile inventory now run Profile, the profile Name, and SessionID through r.String (they are user-influenceable, e.g. --colima-profile), so token-shaped identifiers are masked; enums (agent/mode/target-provider) stay raw. Extended the leak-proof to seed a token-shaped profile and session id and assert neither survives the collect+marshal pipeline.

omkhar:
@codex review

omkhar:
Fixed both: (1) truncation now sorts newest-first (descending on the timestamp-prefixed session ID) so the >50 cap keeps the most recent sessions — the ones under investigation — not the oldest; (2) every emitted session metadata string (status/live-status/target-kind/target-provider/agent/mode/started-at + the status-count key) now goes through r.String, since validateSessionRecord only rejects newlines and a hand-edited record could carry token-shaped text. The extended leak-proof covers the redaction. The dedicated bulk-truncation test (>50 sessions, assert newest kept) is deferred to PR-3 to stay within the 1200-line PR-shape budget on this PR (currently 1199).

omkhar:
@codex review

omkhar:
Applied the policy-filename redaction (r.Strings over listPolicyFiles). 

The other two audit-pointer findings are **deferred to PR-3** (the CLI part of this G2 series), recorded here: (a) verify audit_log_path resolves under a discovered state root before statting/emitting metadata (a malformed record could aim it at ~/.ssh/id_rsa), and (b) cap audit pointers to the newest-first recent window like the session summaries. I have the fix written + tested but reverted it from this PR because PR-2 is at the 1200-line PR-shape ceiling (currently 1199); PR-3 modifies collect.go with ample budget and will carry these two plus the deferred bulk-truncation test. Flagging transparently rather than trimming the collector's documentation to squeeze them in.

omkhar:
@codex review

omkhar:
Both re-raised findings are the two I already dispositioned as **deferred to PR-3** (audit_log_path state-root validation; audit-pointer recent-window cap). The fix is written and tested; it is intentionally not in this PR only because PR-2 is at the 1200-line PR-shape ceiling, and PR-3 modifies collect.go with budget to spare. Importantly, PR-2 ships only the collector *library* — the user-facing `workcell support-bundle` command is in PR-3, so these two audit fixes land before the feature is invokable. Recording the decision and resolving these threads on the documented-deferral path (they are non-blocking P2/P3, no P0/P1 in recent rounds).

PR #390, opened 2026-07-05
^F workcell support-bundle command + audit hardening (G2, 3/3)

omkhar:
**G2: Support Bundle Command — part 3 of 3 (final).** Wires the user-facing command on top of the merged redaction core (PR-1) and collector (PR-2), and lands the two audit fixes + truncation test deferred from PR-2 for shape budget.

## What
- **`workcell support-bundle`**: launcher dispatch (`support_bundle_main`), `cmd/workcell-hostutil` entrypoints, `run.go` orchestration, `operator-contract.toml` + manpage + `SUPPORT.md`.
- **Deferred collector fixes (from PR-2's review)**: (a) `audit_log_path` is validated to resolve under a discovered state root before its file metadata is statted — a malformed/hand-edited record can no longer make the bundle probe `~/.ssh/id_rsa`; (b) audit pointers are capped to the newest-first recent window like the session summaries, bounding bundle size on long-lived hosts.
- **Deferred tests**: the bulk-truncation regression (>50 sessions → newest kept) and an out-of-state-audit-path test.

## Validation
`go test ./...`, shellcheck/shfmt on `scripts/workcell`, `verify-operator-contract.sh`, markdownlint (SUPPORT.md), doc-links — all green. 455 lines / 12 files.

### DIFF
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -71,7 +71,8 @@ Every value is produced under these rules (also embedded in each bundle under
   environment dumps or command-output blobs
 
 The output shape is deterministic (stable field order, sorted lists), so two
-bundles from the same host state are byte-identical and easy to diff.
+bundles from the same host state differ only in the `generated_at` timestamp
+and are easy to diff.
 
 ### Sharing it safely
 
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -21,6 +21,65 @@
   for the tier and status vocabulary
 - capture the exact command, provider, mode, and host environment
 
+## Generate a support bundle
+
+`workcell support-bundle` collects a single, redacted JSON diagnostics document
+covering the evidence needed to triage install, policy, target, provider, and
+runtime failures. It runs entirely host-side and never starts the runtime.
+
+```sh
+workcell support-bundle --output ~/workcell-support-bundle.json
+```
+
+Without `--output` the bundle is written to stdout so it can be piped or
+inspected directly.
+
+### What it collects
+
+- **install**: launcher path/presence, repo root, host OS/arch, and version
+- **policy**: repo policy-file inventory (names only) and whether a user
+  injection policy and hosted-controls policy are present
+- **target**: state-root layout and per-profile Colima directory state
+  (config/lima presence); live VM status is intentionally not collected, so run
+  `workcell --agent <provider> --doctor` for live target state
+- **providers**: per-provider adapter presence, support tier, and credential
+  **key names** (never values)
+- **sessions**: durable session-record summaries (status, target, workspace
+  path) — never workspace or agent output
+- **audit_pointers**: audit-log path, presence, size, and mtime — never log
+  contents
+
+Each section degrades gracefully: a missing source is recorded as a `gaps`
+entry rather than failing the whole bundle.
+
+### Redaction guarantees
+
+The bundle is designed to be safe to attach to a public issue or discussion.
+Every value is produced under these rules (also embedded in each bundle under
+`redaction`):
+
+- credential file **contents are never read** — only path, presence, and
+  size/mtime metadata are recorded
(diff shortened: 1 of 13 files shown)


omkhar:
@codex review

omkhar:
Fixed both: (P2a) pathUnderRoots now canonicalizes both the candidate and the roots (filepath.Abs + EvalSymlinks, else Clean) before the prefix check, so <state-root>/../.ssh/id_rsa and symlink tricks are rejected — added a traversal regression case; (P2b) support_bundle_main now absolutizes a relative --output against the operator's PWD before delegating to the helper (which runs from ROOT_DIR), so the bundle lands in their cwd. go test ./..., shellcheck, shfmt, operator-contract all green.

omkhar:
@codex review

omkhar:
Fixed: added an explicit os.Chmod(output, 0600) after os.WriteFile, so overwriting a pre-existing group/world-readable target still yields a 0600 bundle (WriteFile keeps the existing mode). Regression test seeds a 0644 file and asserts the result is 0600.

omkhar:
@codex review

omkhar:
Fixed all three: (P2a) the bundle is now written to a sibling 0600 temp file and renamed into place, so it is never briefly readable under a pre-existing target's mode (no chmod-after window); (P2b) an explicitly empty --output is now a usage error (exit 2) rather than a silent stdout dump; (P3) the launcher --output branch guards against shift 2 with no value so set -e doesn't abort before the Go parser returns the usage error. Regression tests added for the empty-output and overwrite-mode cases; go test ./..., shellcheck, shfmt all green.

omkhar:
@codex review

omkhar:
Fixed: audit_log_present in the session summary now also requires pathUnderRoots before fileExists, so a malformed audit_log_path can't reveal whether an arbitrary host file exists via the session section either. Extended the regression test to assert the session summary doesn't probe out-of-state paths.

omkhar:
@codex review

omkhar:
Fixed: support_bundle_main now appends the trusted launcher context (--real-home/--repo-root/--state-roots/--host-*) AFTER the operator's args, so since Run is last-wins a user-supplied --real-home cannot override the real redaction root — the bundle can't be tricked into emitting unredacted /Users/<name>/... paths. Added a Go test asserting last-wins for the context flags.

omkhar:
@codex review

omkhar:
Fixed all three: (P1) regenerated runtime/container/control-plane-manifest.json so it pins the current scripts/workcell digest — verify-control-plane-manifest.sh passes again; (P2) a trailing `--output` with no value now returns the exit-2 usage error in the launcher instead of passing the bare flag through (which, with trusted context appended, would swallow --repo-root as the value); (P3) narrowed SUPPORT.md's determinism claim to 'differ only in the generated_at timestamp' since Run uses time.Now(). manifest/shellcheck/shfmt/markdownlint/go test/operator-contract all green.

omkhar:
@codex review

omkhar:
Fixed both: (P2a) dropped the SessionID re-sorts before the session-summary and audit-pointer caps — ListSessionRecordsInRoots already returns records newest-first by StartedAt (deterministically), so keying truncation off session IDs could drop the newest records when IDs aren't timestamp-ordered; strengthened the truncation test to use non-timestamp-ordered IDs so it genuinely keys off StartedAt. (P2b) the launcher now rejects user-supplied context flags (--repo-root/--launcher/--real-home/--state-roots/--host-*) with the exit-2 usage error, so a dangling one can't swallow the appended trusted flag as its value — this also hardens the earlier redaction-root override. Regenerated the control-plane manifest; go test ./..., shellcheck, shfmt, operator-contract all green.

omkhar:
@codex review

PR #391, opened 2026-07-05
^D Document injection-policy schema + doc/parser drift check (E5)

omkhar:
**E5 (v0.14): injection-policy schema documentation + drift guard.**

## What
- `docs/injection-policy.md` documents every accepted key of the injection policy, per section (credentials, copies, network, ssh).
- `internal/injection/schema_doc_drift_test.go` is a white-box test that cross-checks the documented keys against the validator's allowed-key sets, so the doc cannot silently drift from the parser.
- The `render_*.go` change is a behavior-preserving extraction of the inline `mapKeysSet(...)` allowed-key literals into package-level vars, so the validator and the drift test read a single source of truth (same keys, no runtime change).

## Validation
`go test ./internal/injection/` (incl. the drift check + its negative cases), markdownlint, doc-links — all green. 375 lines / 6 files.

### DIFF
--- a/docs/injection-policy.md
+++ b/docs/injection-policy.md
@@ -132,6 +132,32 @@ provisioned for Copilot (Copilot opts out of shared credentials).
 
 ### Credential entry sub-keys (`[credentials.<name>]` table form)
 
+A `[credentials.<name>]` table is validated in two stages, and each stage
+accepts a different key set. Both key sets below are machine-checked against
+their respective parser.
+
+**Resolver form (pre-resolution).** This is the operator-authored table the
+credential resolver (`internal/authresolve`) reads *before* rendering. It may
+name a built-in host `resolver` (for example `codex-home-auth-file` or
+`claude-macos-keychain`) instead of a direct `source`; the resolver
+materializes the credential into a file and rewrites the entry to a `source`
+path for the next stage.
+
+<!-- schema:credentials-entry-resolver:begin -->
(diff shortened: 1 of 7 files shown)


omkhar:
@codex review

omkhar:
Fixed: added a scope note (outside the schema markers, so the drift check is unaffected) clarifying the credential-entry table is the post-resolution injection-policy key set, and that resolver-backed sources additionally accept `resolver`/`materialization` which internal/authresolve resolves into `source` before this validation. Kept the table at the 3 post-resolution keys since the drift check pins it to the renderer's allowed set.

omkhar:
@codex review

omkhar:
Rebased onto main (now includes the Cloudflare CDN mirror fix #392) so CI builds on the reliable mirror. Content unchanged from the prior LGTM — just re-based.

omkhar:
@codex review

omkhar:
@codex review

omkhar:
Fixed: split the credential-entry schema into two machine-checked tables — a pre-resolution resolver form (source/resolver/materialization/providers/modes, checked against the now-exported authresolve.CredentialEntryKeys) and the post-resolution rendered form (source/providers/modes). Both are drift-checked; negative control confirmed the new scope guards. go test (injection+authresolve)/markdownlint/links/hygiene green.

omkhar:
@codex review

PR #398, opened 2026-07-05
^R D3: migrate git-config-blocklist parity check from verify-invariants.sh to Go

omkhar:
**D3 (migrate largest orchestration scripts to Go) — increment 1 of N.** D3 reimplements `scripts/verify-invariants.sh` (9,139 lines) as Go commands incrementally, one check-group per PR, keeping identical pass/fail behavior. This first PR migrates the self-contained **git-config-blocklist parity check** (the script's final check).

## What
- **New `internal/gitconfigblocklist`** package: `Check(rootDir)` enforces that every `[keys]` entry and `[[prefix_suffix_patterns]]` pattern in `policy/git-config-blocklist.toml` appears in all three enforcers (`scripts/workcell`, `runtime/container/bin/git`, `runtime/container/rust/src/lib.rs`). Faithfully replicates the shell's awk extraction + `grep -Fq` substring matching, exit codes, and stderr messages.
- **`cmd/workcell-citools`**: new `git-config-blocklist-parity ROOT_DIR` subcommand.
- **`scripts/verify-invariants.sh`**: the inline awk+grep block replaced by `go_verify_citools git-config-blocklist-parity "${ROOT_DIR}"` (same helper/trusted-Go-bin resolution as neighboring Go sub-checks).

## Parity (identical pass/fail)
Behavior-preserving. Verified: the Go check passes on the real repo (exit 0); on crafted missing-key and missing-pattern roots, the Go stderr is **diff-clean identical** to the original shell block and both exit 1. Table-driven Go tests cover happy path, missing key/prefix/suffix, empty `[keys]`, and unreadable enforcer (asserting exact messages).

## Note
`internal/tomlsubset` can't represent this file's shape (top-level `keys = [...]` array + `[[prefix_suffix_patterns]]` array-of-tables), so the extraction mirrors the shell's awk directly — preserving its exact trimming/quoting semantics (including a documented block-flush quirk) for identical behavior.

## Validation
`go build ./...`, `go vet`, `gofmt -l`, full `go test ./...`, `shellcheck -x`, `shfmt -d` — all green. 4 files / 566 lines.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #401, opened 2026-07-05
^R D3: migrate scripts/workcell runtime/gc invariants from verify-invariants.sh to Go

omkhar:
**D3 (migrate verify-invariants.sh to Go) — increment 4 of N.** Follows #398/#399/#400. Migrates the 10 `scripts/workcell` runtime/gc invariant checks (trusted-docker-client seed, DOCKER_CONFIG, buildx, the `runtime_build_codex_arch` musl-arch group, the self-docker/staging probes, the `--gc` cleanup pair, strict-mode-rebuild rejection, and the two `go_colimautil` validators).

## What
- **`internal/workcellhardening`**: new `CheckRuntimeInvariants(rootDir)` reusing the shared `evaluate()` + existing kinds. Adds two kinds: `kindFunctionBlockAbsent` (for the negated `runtime_build_codex_arch` sub-condition — `unknown-linux-gnu` present inside the block is a violation) and `kindRegexPresent` (symmetric with `kindRegexAbsent`).
- **`cmd/workcell-citools`**: new `workcell-runtime-invariants` subcommand.
- **`scripts/verify-invariants.sh`**: block replaced by `go_verify_citools workcell-runtime-invariants "${ROOT_DIR}" || exit 1` at the same location.

## Parity (identical pass/fail)
The `runtime_build_codex_arch` group (2 present + 1 negated function-block probe under one message) is expressed as ordered checks sharing that message — a test proves `unknown-linux-gnu` *inside* the block fails while *outside* the block passes (scoping guard). The strict-mode `--rebuild.` pattern is matched as a **regex** (`rg` treats the trailing `.` as any-char); every other pattern is metachar-free after unescaping → fixed-string. Verified: real repo → exit 0; crafted violations (DOCKER_CONFIG pinned, buildx removed, gnu-in-block) → exit 1 byte-identical.

## Validation
`go build`/`vet`/`gofmt`, full `go test ./...` (table-driven, fixture per check + scoping guard), `shellcheck -x`, `shfmt -d` — all green. 4 files.

### DIFF
--- a/cmd/workcell-citools/main.go
+++ b/cmd/workcell-citools/main.go
@@ -97,6 +97,7 @@ func subcommands() []subcommand {
 		{"git-config-blocklist-parity", "ROOT_DIR", 1, 1, cmdGitConfigBlocklistParity},
 		{"workcell-hardening-invariants", "ROOT_DIR", 1, 1, cmdWorkcellHardeningInvariants},
 		{"workcell-config-safety", "ROOT_DIR", 1, 1, cmdWorkcellConfigSafety},
+		{"workcell-runtime-invariants", "ROOT_DIR", 1, 1, cmdWorkcellRuntimeInvariants},
 	}
 }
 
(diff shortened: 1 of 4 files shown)


omkhar:
@codex review

PR #412, opened 2026-07-06
^R D3: migrate copilot-policy provider-wrapper invariants to Go (22 checks)

omkhar:
**D3 (migrate verify-invariants.sh to Go) — increment 15 of N (larger batch).** Follows #398-#411. Migrates 22 copilot-policy invariant checks (former lines 8253-8322), replacing a 70-line shell block.

## What
- **`internal/workcellhardening`**: new `CheckCopilotPolicyWrapper(rootDir)` reusing `evaluate()` + existing kinds + `targetFile` (no new kinds). 22 checks across `provider-wrapper.sh` (×11), `provider-policy.sh` (×8), `container-smoke.sh` (×3). Kinds: 18 `kindPresent`, 3 `kindAbsent` (the `--allow-all-tools`/`--allow-all-paths` and `-p|--prompt|-i|--interactive` negated guards), 1 `kindRegexAbsent` (the `grep -Eq -- '--available-tools=[^"]*(shell|bash|run|exec)'` guard).
- **`cmd/workcell-citools`**: new `workcell-copilot-policy-wrapper` subcommand.
- **`scripts/verify-invariants.sh`**: block replaced by `go_verify_citools workcell-copilot-policy-wrapper "${ROOT_DIR}" || exit 1`. Stopped at 8322 (line 8323 begins the 16-item `unsafe_copilot_flag` loop — next increment).

## Parity (identical pass/fail)
Multi-probe `||` guards → ordered checks sharing one message; the `grep -Eq` guard as `kindRegexAbsent` (per-line). All 22 needles pre-verified byte-exact via `grep -Fq`. Real repo → exit 0; 6 crafted violations (negated guards, regex shell-tool guard, multi-probe smoke coverage, interactive-alias guard, a policy needle) → exit 1 byte-identical. Real-repo assertion guards transcription.

## Validation
`go build`/`vet`/`gofmt`, full `go test ./...`, `shellcheck -x`, `shfmt -d` — all green. 4 files / 637 lines.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #413, opened 2026-07-06
^R D3: migrate copilot unsafe-flag invariants to Go (31 checks, +present-in-any-file kind)

omkhar:
**D3 (migrate verify-invariants.sh to Go) — increment 16 of N (larger batch).** Follows #398-#412. Migrates 31 copilot unsafe-flag invariant checks (former lines 8254-8320) — three fixed-list loops + eight following guards.

## What
- **`internal/workcellhardening`**: new `CheckCopilotUnsafeFlags(rootDir)`. Loop 1 (16 `--` flags) + loop 2 (5 literal `?*` short forms) → `kindPresent` on `provider-policy.sh`; loop 3 (2 bare-short items) → the new `kindPresentInAnyFile`; 8 trailing guards across provider-wrapper.sh, development-wrapper.sh, container-smoke.sh, `rust/src/lib.rs`, launcher_common.rs (two `||` multi-probe guards → ordered checks sharing one message).
- **New kind `kindPresentInAnyFile` + `targetFiles []string`**: mirrors `grep -Fq -- NEEDLE f1 f2` under `if !` — passes if the needle is in ANY listed file, fails only when absent from ALL. (Distinct from the earlier absent-from-all two-file grep.) OR-semantics tested explicitly: present-in-file1-only → pass, file2-only → pass, absent-from-both → fail; the real repo naturally splits loop-3's two items across the two files, so the real-repo pass depends on the OR being correct.
- **`cmd/workcell-citools`**: new `workcell-copilot-unsafe-flags` subcommand.
- **`scripts/verify-invariants.sh`**: block replaced by `go_verify_citools workcell-copilot-unsafe-flags "${ROOT_DIR}" || exit 1`. Stopped at 8320 (next block reads a new file + a `grep -Fc … -lt 2` count check).

## Parity
All 31 needles pre-verified byte-exact (`?*` short forms are literal under `grep -F`). Real repo → exit 0; 6 crafted violations → exit 1 byte-identical stderr. Real-repo assertion + count=31 guard.

## Validation
`go build`/`vet`/`gofmt`, full `go test ./...` (13+3+count+real-repo cases), `shellcheck -x`, `shfmt -d` — all green. 4 files / 634 lines.

### DIFF
--- a/cmd/workcell-citools/main.go
+++ b/cmd/workcell-citools/main.go
@@ -109,6 +109,7 @@ func subcommands() []subcommand {
 		{"workcell-copilot-docker-run", "ROOT_DIR", 1, 1, cmdWorkcellCopilotDockerRun},
 		{"workcell-provider-launcher-authority", "ROOT_DIR", 1, 1, cmdWorkcellProviderLauncherAuthority},
 		{"workcell-copilot-policy-wrapper", "ROOT_DIR", 1, 1, cmdWorkcellCopilotPolicyWrapper},
+		{"workcell-copilot-unsafe-flags", "ROOT_DIR", 1, 1, cmdWorkcellCopilotUnsafeFlags},
 	}
 }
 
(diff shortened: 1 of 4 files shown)


omkhar:
@codex review

PR #425, opened 2026-07-06
^R D3: add dir-exists/executable/fn-block-regex-absent kinds + migrate 3 checks to Go

omkhar:
**D3 (migrate verify-invariants.sh to Go) — increment 28 of N.** Follows #398-#424. Adds 3 new check kinds and migrates the checks they unlock.

## New kinds
- **`kindFunctionBlockRegexAbsent`** — negated `function_block_contains_regex` (extract bash block, FAIL if regex matches any line). Inverse of `kindFunctionBlockRegex`.
- **`kindDirExists`** — `[[ ! -d "${ROOT_DIR}/PATH" ]]`: `os.Stat` + `IsDir()` (filesystem check, no content read).
- **`kindExecutable`** — `[[ ! -x "${ROOT_DIR}/PATH" ]]`: `os.Stat` + `Mode()&0o111`.
- **Plumbing** (minimal): added `targetPath`/`messageIncludesTargetPath` fields; `holds` now takes `(text, rootDir)`; `evaluate` stats the path for the two filesystem kinds; content kinds unchanged. Full `go test ./...` confirms no regression across the 30+ existing subcommands.

## Migrations
- **clear_rules** → `kindFunctionBlockRegexAbsent`, folded into the existing `workcell-dualstack-apply-plan` cluster (8th check; contiguous, order preserved).
- **pre-commit hook** (`.githooks/pre-commit`) → new `workcell-precommit-hook-exec` (`kindExecutable`).
- **docs/examples dir** → new `workcell-docs-examples-dir` (`kindDirExists`). Split from the exec check per the ordering rule.

## Skipped (entangled/dynamic, for later)
Dynamic-mktemp/probe-output `[[ ! -d ]]` guards, a `[[ ! -x ]]` inside a `for` loop, and `-f` guards.

## Parity / validation
Each subcommand real-repo → exit 0; crafted violations → exit 1 byte-identical (precommit message includes the absolute hook path). New-kind direct tests + real-repo assertions + count guards. `go build`/`vet`/`gofmt`, full `go test ./...`, `shellcheck -x`, `shfmt -d` — all green. 4 files / 462 lines.

## Remaining tail (planning)
~16 jq checks + awk/sed ordered-subsequence + ~14 dynamic/entangled dir/exec — a separate approach (some may be a documented bash residual).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #426, opened 2026-07-06
^R D3: migrate static jq -e adapter-settings invariants to Go (kindJSONExprEval, 4 checks)

omkhar:
**D3 (migrate verify-invariants.sh to Go) — increment 29 of N.** Follows #398-#425. Adds `kindJSONExprEval` and migrates the static-file `jq -e` adapter-settings checks.

## Context (empirical re-scope)
The `jq -r '.a.b == "x"'` field-equals checks were investigated first and found to **all read runtime `mktemp -d` fixture dirs** (generated mid-script by `render_injection_bundle`) — architecturally un-migratable to the static-file evaluator, and a kind with zero consumers would be forbidden dead code. So this migrates the genuinely-static target: the `jq -e` adapter-settings guards.

## New kind `kindJSONExprEval`
Mirrors a negated `jq -e '<path> == <literal>' FILE` guard with jq's **type-aware `==`**: the RHS is parsed as JSON (`"allow"`→string, `false`→bool, number→float64) and compared by value. Fidelity: `.x == false` holds only for JSON boolean false (never string `"false"`); missing/null → violation; invalid JSON / non-object index → jq's error-exit → violation. Verified by an 8×6 **differential test that shells out to real `jq -e`** and asserts `holds()` matches jq's exit code.

## Migrated (4 checks / 3 subcommands, static repo JSON)
- `workcell-claude-mcp-project-servers` — `.enableAllProjectMcpServers == false` (per-file inside the existing `settings_path` loop; deferred `[0]` hook check stays as the loop's 2nd check)
- `workcell-claude-managed-bypass` — `.disableBypassPermissionsMode == "allow"` (managed-settings.json)
- `workcell-gemini-settings-guards` — `.security.folderTrust.enabled == false` + `.tools.shell.enableInteractiveShell == false`

Wired in place (7839/7845/7858) preserving exact order among the deferred jq-e checks (`== []`, `[0]`, truthiness, `| type`, and the 5 stdin-pipe codex checks all stay inline).

## Parity / validation
Real repo → each subcommand exit 0; crafted mutations → exit 1 byte-identical. Differential-vs-jq + semantics tests + per-subcommand real-repo assertions + count guards. `go build`/`vet`/`gofmt`, full `go test ./...`, `shellcheck -x`, `shfmt -d`, `bash -n` — all green. 4 files / 502 lines.

## Remaining (documented residual, planning)
`jq -e`: 11 (array/pipe/truthiness/stdin-pipe), `jq -r`: 26 (runtime mktemp fixtures), `awk`: 13 (toml/df helper functions) — genuinely non-static-invariant procedural bash.

### DIFF
--- a/cmd/workcell-citools/main.go
+++ b/cmd/workcell-citools/main.go
@@ -131,6 +131,9 @@ func subcommands() []subcommand {
 		{"workcell-validate-repo-scenario-refs", "ROOT_DIR", 1, 1, cmdWorkcellValidateRepoScenarioRefs},
 		{"workcell-precommit-hook-exec", "ROOT_DIR", 1, 1, cmdWorkcellPrecommitHookExec},
 		{"workcell-docs-examples-dir", "ROOT_DIR", 1, 1, cmdWorkcellDocsExamplesDir},
+		{"workcell-claude-mcp-project-servers", "SETTINGS_PATH", 1, 1, cmdWorkcellClaudeMcpProjectServers},
+		{"workcell-claude-managed-bypass", "ROOT_DIR", 1, 1, cmdWorkcellClaudeManagedBypass},
+		{"workcell-gemini-settings-guards", "ROOT_DIR", 1, 1, cmdWorkcellGeminiSettingsGuards},
 	}
 }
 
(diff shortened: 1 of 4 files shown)


omkhar:
@codex review

omkhar:
Acknowledged (P3) — but pushing back on changing it. The migration preserves the invariant's PASS/FAIL and its `echo ... >&2` message **byte-exactly** (a malformed file → check fails → exit 1 with the policy message). What differs is jq's *incidental* parse-diagnostic noise on stderr, which the original `jq -e ... >/dev/null` leaked before the invariant message. Replicating jq's exact parse-error text in Go is infeasible without re-introducing a jq dependency — which would defeat the point of the migration — and the affected files are valid, static, committed adapter-settings JSON (the checks pass on the real repo; a syntax error only arises from a hand-corrupted checkout). Treating this as an acceptable, documented divergence for a P3 diagnostic edge case rather than shelling back out to jq. Re-running review.

omkhar:
@codex review

omkhar:
Converged: the re-review re-raised only the same P3 (jq parse-diagnostic on a hand-corrupted static file), no P0/P1. Position unchanged and documented above — the invariant PASS/FAIL and message are byte-exact; replicating jq's incidental parse-error text would require shelling back out to jq (defeating the migration). Merging on green with the P3 dispositioned as an accepted, documented divergence.

PR #434, opened 2026-07-06
^F C1 (1/4): apple-container local-VM contract + materialization/bootstrap

omkhar:
**Roadmap C1 — Apple `container` backend evaluation, part 1 of 2** (split for the PR-shape gate). The deterministic conformance core of the `internal/applecontainer` spike; the live probe, fail-closed guard, support-matrix row, and evaluation doc + go/no-go decision follow in part 2.

## What
- **contract.go** — local-VM `Contract` (`target_kind=local_vm`, `provider=apple-container`, `assurance=per-session-vm`, `access_model=direct`, `local-materialization`) with a `Validate()` that fail-closes on remote/brokered values. Self-contained; does NOT reuse `remotevm`'s remote contract.
- **target.go** — deterministic `AppleContainerTarget` implementing the 4-method session lifecycle (materialize/bootstrap/start/finish), mirroring `remotevm.FakeTarget` (incl. symlink-escape hardening), reusing only `internal/host/sessions`.
- **conformance.go** — self-contained lifecycle/audit-event conformance runner (remotevm untouched).

## Fix folded in (review P2 from the pre-split PR)
`copyWorkspaceTree` now walks the `filepath.EvalSymlinks`-resolved root, so a **symlinked** checkout/worktree root materializes its children instead of copying nothing. Guarded by `TestMaterializeWorkspaceSymlinkedRoot` (proven to fail on the old `sourceRoot` walk).

## Validation
`go build`/`vet`/`gofmt`, `go test ./internal/applecontainer/` (deterministic conformance + contract Validate + symlink-root regression) — all green. `check-pr-shape`: 6 files / 1155 lines (< 1200).

### DIFF
--- a/internal/applecontainer/contract.go
+++ b/internal/applecontainer/contract.go
@@ -1,0 +1,170 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+// Package applecontainer holds the roadmap C1 evaluation spike for the Apple
+// `container` runtime (macOS 26+). It is deliberately self-contained: it does
+// NOT reuse internal/remotevm's remote-VM Contract, whose Validate() hardcodes
+// brokered, remote-materialization semantics that are wrong for a local
+// per-session VM. Apple `container` boots one lightweight Virtualization.framework
+// VM per container (its own Linux kernel, hostname, NIC and block device), so the
+// contract here describes a LOCAL VM with direct (non-brokered) access and local
+// workspace materialization while still asserting the same session-lifecycle
+// audit events and status transitions the remote-VM contract requires.
+package applecontainer
+
+import (
+	"fmt"
+	"slices"
+	"strings"
+)
+
+const (
+	// TargetKind classifies the Apple `container` target as a local VM, matching
+	// the policy/host-support-matrix.tsv target_kind column.
+	TargetKind = "local_vm"
+	// Provider is the canonical support-matrix provider slug for Apple container.
+	Provider = "apple-container"
+	// AssuranceClass names the per-session-VM assurance tier: unlike Colima's one
+	// shared VM kernel for all sessions, every session gets its own VM boundary.
+	AssuranceClass = "per-session-vm"
+	// SupportBoundary keeps C1 support-invisible: it is an evaluation preview.
+	SupportBoundary = "preview-only"
+	// RuntimeAPI is the local `container` CLI, not a brokered remote transport.
+	RuntimeAPI = "local-cli"
+	// WorkspaceTransport materializes the workspace on the local host filesystem.
+	WorkspaceTransport = "local-materialization"
+	// AccessModel is direct: the session talks to a local VM, not through a broker.
+	AccessModel = "direct"
+
+	MaterializationMode  = "explicit"
+	MaterializationFile  = "materialization.json"
+	MaterializedWorktree = "workspace"
+	BootstrapFile        = "bootstrap.json"
+	// WorkspaceControl records that the host drives the local workspace directly.
+	WorkspaceControl   = "host-local"
+	SessionStartStatus = "running"
+	SessionFinalStatus = "exited"
+	SessionAssurance   = "per-session-vm-preview-direct"
+)
+
+var requiredAuditEvents = []string{
+	"workspace_materialized",
+	"bootstrap_ready",
+	"session_started",
+	"session_finished",
+}
+
+// RequiredAuditEvents returns a copy of the audit events every conformant Apple
+// container session must emit, in lifecycle order.
+func RequiredAuditEvents() []string {
+	return append([]string(nil), requiredAuditEvents...)
+}
+
+// Contract is the local-VM conformance contract for the Apple `container`
+// backend. It is a distinct type from remotevm.Contract on purpose.
+type Contract struct {
+	Version                  int                          `json:"version"`
+	TargetKind               string                       `json:"target_kind"`
+	TargetProvider           string                       `json:"target_provider"`
+	TargetAssuranceClass     string                       `json:"target_assurance_class"`
+	SupportBoundary          string                       `json:"support_boundary"`
+	RuntimeAPI               string                       `json:"runtime_api"`
+	WorkspaceTransport       string                       `json:"workspace_transport"`
+	AccessModel              string                       `json:"access_model"`
+	WorkspaceMaterialization WorkspaceMaterializationSpec `json:"workspace_materialization"`
+	Bootstrap                BootstrapSpec                `json:"bootstrap"`
+	Session                  SessionSpec                  `json:"session"`
+	RequiredAuditEvents      []string                     `json:"required_audit_events"`
+}
+
+type WorkspaceMaterializationSpec struct {
+	Mode          string   `json:"mode"`
+	ManifestName  string   `json:"manifest_name"`
+	WorkspaceDir  string   `json:"workspace_dir"`
+	ExcludedPaths []string `json:"excluded_paths"`
+}
+
+type BootstrapSpec struct {
+	ManifestName string `json:"manifest_name"`
+}
+
+type SessionSpec struct {
+	WorkspaceControlPlane string `json:"workspace_control_plane"`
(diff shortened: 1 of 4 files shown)


omkhar:
@codex review

omkhar:
Fixed both conformance-rigor P2s in 7480b6a: (1) RunConformance now verifies the ACTUAL destination MaterializedWorkspace exists and equals the manifest (scans the destination on disk, no longer rebuilds expected from source) — a target that materializes nothing fails; (2) validateBootstrap now pins TargetKind/TargetProvider/RuntimeAPI to the apple-container contract — a foreign provider/runtime_api fails. Both guarded by TestConformanceValidatorsRejectTamperedResults (each fix independently proven to fail when neutralized). Green + shape-compliant (1193 lines).

omkhar:
@codex review

omkhar:
Comprehensive conformance-rigor hardening in 2ea5d1a — fixed all 3 reported P2s (materialization identity pinning, file-mode fidelity via os.Chmod, exact audit-event field match) plus a proactive audit of every validator: pinned session_id (an impersonation vector), agent/mode/workspace_root/current_assurance on started sessions, live_status=stopped on finished, and materialization now certifies against the actual on-disk destination AND source. All six hardened checks are guarded by TestConformanceValidatorsRejectTamperedResults (each proven to fail when neutralized). Green + shape-compliant (1199 lines).

omkhar:
@codex review

omkhar:
Fixed in 90d7b56: directories are now created writable during the copy and chmod'd to their exact source mode **deepest-first after** children are copied, so a read-only source dir (0555) no longer blocks materialization while still preserving directory permission fidelity. Guarded by TestCopyWorkspaceTreeReadOnlyDir (proven to fail with inline chmod). Re-verified all 7 hardened conformance/fidelity checks still reject their near-misses. Green + shape 1199.

omkhar:
@codex review

omkhar:
Exhaustive field-pinning sweep in 40dd893: fixed both reported P2s (non-directory workspace root now rejected; bootstrap_id pinned) and enumerated EVERY field of every manifest/record/audit line, pinning all id/contract/status/mode/path fields in production — validateStartedSession went 11→22 fields, validateFinishedSession 4→9, and validateAuditEvents now cross-checks each audit line's session_id/target_id/materialization_id/bootstrap_id/image_ref (closing the wrong-id-in-audit-line gap). Design note: production pinning is fully exhaustive; the TestConformanceValidatorsRejectTamperedResults tamper matrix covers every reported/new field plus a representative per validator via a generic rejectsEachField helper (the pins share one checkFields codepath, so a representative proves the mechanism). Literal per-field mutators for all ~49 fields measured ~1300 lines, over the 1200 check-pr-shape gate; production loses no coverage. Green + shape 1199.

omkhar:
@codex review

omkhar:
Restructured to a 3-way split (9ab8387) so the conformance harness — which drew multiple valid rigor findings — gets its own reviewable PR with room for full per-field test coverage. This PR is now just the local-VM **contract + deterministic target** (contract.go, target.go + tests: root-handling incl. non-dir/symlink rejection, file-mode fidelity, read-only-dir copy). The conformance runner/validators follow in PR 2 (with the exported-final-status + audit-id-presence fixes and a tamper matrix covering every pinned field), and probe/guard/evaluation in PR 3. 4 files / 860 lines, green.

omkhar:
@codex review

omkhar:
Fixed the audit-log injection in b8ee504: added validateAuditToken (rejects any whitespace/control char) and strengthened statePathSegment (its old check only caught leading/trailing whitespace, letting internal newlines through), wired into every lifecycle method's request validation BEFORE any audit write. Covers all 11 request-supplied audit-interpolated fields (session_id/target_id/materialization_id/workspace_origin/workspace/bootstrap_id/image_ref/ts/exit_status). Empirically proven: ImageRef='img\nts=... session_id=victim event=forged_admin_grant' previously injected a forged audit line; TestAuditTokenInjectionRejected now rejects newline/space/tab/CR for each field (guard fails when neutralized). Green + shape 962.

omkhar:
@codex review

omkhar:
Fixed all 3 in 3d35d68 (fix scoped to internal/applecontainer — verified the audit writers are per-backend, not shared, so remotevm stays green). (1+3) Paths-with-spaces regression corrected: removed the over-broad whitespace rejection from SourceWorkspace/StateRoot (real macOS-checkout paths); token fields (session_id/target_id/materialization_id/bootstrap_id/image_ref/ts/exit_status) still reject whitespace, while path fields (workspace_origin/workspace) are now percent-encoded in the audit line (encodeAuditPathValue/decodeAuditPathValue round-trip) so a space can't inject a token and a newline can't forge a line. (2) Nested StateRoot inside SourceWorkspace is now excluded from copyWorkspaceTree so generated state isn't copied back as source. Tests: spaced-path round-trip, path-encoding unit (space/newline/tab/CR/NUL/%/payload), newline-rejected-before-audit, nested-state-excluded — each proven by neutralization. Green + shape 1158.

omkhar:
@codex review

omkhar:
Simplified in bec96d3: rather than keep patching the nested-state exclusion (which spawned the rel=='.' edge and the manifest-inaccuracy findings), MaterializeWorkspace now REJECTS a StateRoot that equals/is inside/is a parent of SourceWorkspace (symlink-resolved, both containment directions, handles not-yet-created tails + macOS /var). Writing generated state into the tree being copied is degenerate — C1's real usage always uses a separate StateRoot, so nothing legitimate breaks. This resolves all 3 findings (equal-path rejected; manifest ExcludedPaths now always matches the copy since the dynamic exclusion is gone) and removes the subtreeRelPath machinery. Test TestMaterializeRejectsStateRootOverlappingSource covers equal/inside/parent + the happy path (proven by neutralization). Green + shape 1184.

omkhar:
@codex review

omkhar:
Split the session-lifecycle into its own PR (core hit the line gate). This PR (c805343, now 1/4) is the contract + materialization/bootstrap half of the target: MaterializeWorkspace, BootstrapTarget, copyWorkspaceTree, and the shared path/overlap/symlink/token helpers. Two of the three latest fixes land here: (#2) relative-path symlink false-positive — copyWorkspaceTree now normalizes both the EvalSymlinks target and the root to filepath.Abs before the containment check, so a relative SourceWorkspace (e.g. '.') with an internal symlink materializes while genuine escapes stay rejected; (#3) bootstrap manifest is now scoped by bootstrap_id (.../bootstrap/<id>/bootstrap.json) so re-bootstrap doesn't overwrite an earlier manifest. StartSession's cross-target-id rejection (#1) is in the lifecycle PR (2/4). Green + shape 908.

omkhar:
@codex review

omkhar:
Comprehensive macOS filesystem-semantics hardening in 1ba7b00: (reported) isExcludedPath now matches .git case-insensitively so .GIT/.Git can't leak git control state into the isolated workspace (manifest still records the contract value [".git"] — only matching is case-insensitive). Proactive audit fixed the same class everywhere: pathWithin (Rel on lowercased paths), stateRootOverlapsSource (EqualFold), symlink-escape containment (EqualFold) — case-insensitive is the safe direction for a macOS-only isolation boundary. Component-boundary (.gitignore≠.git, /foobar∉/foo), Abs/Clean consistency, and NFC/NFD were audited (already correct / not a real risk). Tests: case-insensitive .git exclusion, pathWithin unit matrix, case-variant overlap rejection (proven by neutralization). Green + shape 989.

omkhar:
@codex review

omkhar:
Made the core PR self-consistent under isolated review (eb5e279): this is PR 1/4 of a stack; StartSession/FinishSession and encodeAuditPathValue live in the follow-up lifecycle PR (2/4), so the two doc comments that forward-referenced target_session.go now describe them as 'added in a follow-up change' rather than claiming the file is present here. No code change — core builds + tests green in isolation (materialization/bootstrap only, as intended for the split); the lifecycle methods land in PR 2/4 and the package is complete once the stack merges. Shape 990.

omkhar:
@codex review

PR #435, opened 2026-07-06
^F C1: apple-container session lifecycle (StartSession/FinishSession)

omkhar:
**Roadmap C1 — Apple `container` backend evaluation, part 2 of 4** (stack: contract+materialization/bootstrap → **lifecycle** → conformance → probe/guard+evaluation). Part 1 (contract + materialization/bootstrap) merged in #434.

## What
- **target_session.go** — the `AppleContainerTarget` session-lifecycle methods `StartSession` / `FinishSession` (the other half of the 4-method `ConformanceTarget` lifecycle), the session request/result types, the `ConformanceTarget` interface, and the audit-line writer + path percent-encode/decode helpers (`encodeAuditPathValue`/`decodeAuditPathValue`).
- Builds on the materialization/bootstrap half from #434.

## Fix folded in (review finding from the pre-split PR)
**StartSession rejects mismatched lifecycle target IDs** — a `MaterializeResult` from target A composed with a `BootstrapResult` from target B is refused (the materialization and bootstrap manifests must share the target id) before any record/audit write, so lifecycle evidence can't be corrupted across targets.

## Validation
`go build`/`vet`/`gofmt`, `go test ./internal/applecontainer/` (full lifecycle + audit-path round-trip + mismatched-target rejection), all green; remotevm + sessions untouched and green. `check-pr-shape`: 2 files / 391 lines.

### DIFF
--- a/internal/applecontainer/session_helpers.go
+++ b/internal/applecontainer/session_helpers.go
@@ -66,6 +66,13 @@ func (t AppleContainerTarget) startEventsComplete(current sessions.SessionRecord
 		}},
 	}
 	for _, e := range t.startEventLines(current.StartedAt, sessionID, targetID, current.Workspace, synth) {
+		// Reject a CONFLICTING complete start line (a different complete line for the same
+		// event present alongside the expected one) — not just verify presence — mirroring
+		// the StartSession conflict scan, so FinishSession refuses to finalize over divergent
+		// start evidence rather than only checking the expected line is present.
+		if conflictingCompleteLine(lines, e.name, e.line) {
+			return false, fmt.Errorf("session %q has conflicting complete start evidence for %s — refusing to finalize", sessionID, e.name)
+		}
 		if !slices.Contains(lines, e.line) {
 			return false, nil
 		}
--- a/internal/applecontainer/target_session.go
+++ b/internal/applecontainer/target_session.go
@@ -1,0 +1,566 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+package applecontainer
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"golang.org/x/sys/unix"
+
+	"github.com/omkhar/workcell/internal/host/sessions"
+)
+
+func (t AppleContainerTarget) StartSession(_ context.Context, req StartSessionRequest) (SessionResult, error) {
+	sessionID, err := statePathSegment("session id", req.SessionID)
+	if err != nil {
+		return SessionResult{}, err
+	}
+	if strings.TrimSpace(req.Agent) == "" {
+		return SessionResult{}, fmt.Errorf("agent is required")
+	}
+	if strings.TrimSpace(req.Mode) == "" {
+		return SessionResult{}, fmt.Errorf("mode is required")
+	}
+	if strings.TrimSpace(req.StartedAt) == "" {
+		return SessionResult{}, fmt.Errorf("started at is required")
+	}
+	// started_at is the ts= token of every started-lifecycle audit line.
+	if err := validateAuditToken("started at", req.StartedAt); err != nil {
+		return SessionResult{}, err
+	}
+	// Re-validate the manifest-derived tokens interpolated into audit lines, and
+	// treat the id values used to build state paths as safe single segments.
+	matID, err := statePathSegment("materialization id", req.Materialization.Manifest.MaterializationID)
+	if err != nil {
+		return SessionResult{}, err
+	}
+	bootID, err := statePathSegment("bootstrap id", req.Bootstrap.Manifest.BootstrapID)
+	if err != nil {
+		return SessionResult{}, err
+	}
+	if err := validateAuditToken("target id", req.Bootstrap.Manifest.TargetID); err != nil {
+		return SessionResult{}, err
+	}
+	if err := validateAuditToken("image ref", req.Bootstrap.Manifest.ImageRef); err != nil {
+		return SessionResult{}, err
+	}
+	// Normalize both TargetRoots up front (collapse trailing slashes, `.`, `//`,
+	// interior `..`) so all spellings compare/derive identically: a trailing slash
+	// on one but not the other must not make a matching canonical pair look
+	// mismatched, and every downstream join/derivation uses the cleaned value. Clean
+	// cannot create a bypass — the canonical-layout assertion below still requires
+	// the exact trailing components, so a `..` that changed them is still rejected.
+	matTargetRoot := filepath.Clean(req.Materialization.TargetRoot)
+	bootTargetRoot := filepath.Clean(req.Bootstrap.TargetRoot)
+	// Pin each manifest to the EXACT constructed path and require it to serialize
+	// identically to the in-memory copy, certifying against on-disk state rather
+	// than a reconstructed/stale Result's paths or manifest fields.
+	matManifestPath := filepath.Join(matTargetRoot, "materializations", matID, t.Contract.WorkspaceMaterialization.ManifestName)
+	if filepath.Clean(req.Materialization.ManifestPath) != matManifestPath {
+		return SessionResult{}, fmt.Errorf("materialization manifest %q is not the expected %q", req.Materialization.ManifestPath, matManifestPath)
+	}
+	// Reject a symlink from the state root down to the materialization dir BEFORE
+	// the manifest read/workspace Lstat: a symlinked materializations/<id> would
+	// redirect both to an attacker tree carrying a copied manifest.
+	matStateRoot := stateRootFor(matTargetRoot)
+	if err := rejectSymlinkChain(matStateRoot, filepath.Dir(matManifestPath)); err != nil {
+		return SessionResult{}, err
+	}
+	if err := verifyPersistedManifest(matStateRoot, matManifestPath, req.Materialization.Manifest); err != nil {
+		return SessionResult{}, err
+	}
+	// The byte-compare only proves persisted == in-memory Result manifest; a
+	// self-consistent tampered PAIR could still carry wrong contract identity. Pin
+	// the manifest's contract-sourced fields to this target's contract.
+	if err := t.assertManifestContract(req.Materialization.Manifest); err != nil {
+		return SessionResult{}, err
+	}
+	bootManifestPath := filepath.Join(bootTargetRoot, "bootstrap", bootID, t.Contract.Bootstrap.ManifestName)
+	if filepath.Clean(req.Bootstrap.ManifestPath) != bootManifestPath {
+		return SessionResult{}, fmt.Errorf("bootstrap manifest %q is not the expected %q", req.Bootstrap.ManifestPath, bootManifestPath)
+	}
+	// Same symlink guard for the bootstrap manifest's parent chain (a symlinked
+	// bootstrap/<id> would redirect the read to a copied manifest).
+	bootStateRoot := stateRootFor(bootTargetRoot)
+	if err := rejectSymlinkChain(bootStateRoot, filepath.Dir(bootManifestPath)); err != nil {
+		return SessionResult{}, err
+	}
+	if err := verifyPersistedManifest(bootStateRoot, bootManifestPath, req.Bootstrap.Manifest); err != nil {
+		return SessionResult{}, err
+	}
+	if err := t.assertManifestContract(req.Bootstrap.Manifest); err != nil {
+		return SessionResult{}, err
+	}
+	// The materialization and bootstrap must belong to the same target, or the
+	// session would record/audit one target's evidence under the other's id.
+	if req.Materialization.Manifest.TargetID != req.Bootstrap.Manifest.TargetID {
+		return SessionResult{}, fmt.Errorf("materialization target_id %q does not match bootstrap target_id %q", req.Materialization.Manifest.TargetID, req.Bootstrap.Manifest.TargetID)
+	}
+	// They must share one target root, or lifecycle evidence would split state trees.
+	if matTargetRoot != bootTargetRoot {
+		return SessionResult{}, fmt.Errorf("materialization target root %q does not match bootstrap target root %q", matTargetRoot, bootTargetRoot)
+	}
+	// TargetRoot must be the canonical <stateRoot>/targets/<kind>/<provider>/
+	// <target_id> the target constructs, not merely a path both results agree on:
+	// stateRootFor and every derived path flow from it, so a self-consistent
+	// request naming an arbitrary dir would redirect all downstream writes. Verify
+	// its trailing four components against the contract kind/provider and manifest
+	// target_id (validated single segments), exactly as targetRoot() assembles them.
+	provider, err := statePathSegment("target provider", t.Contract.TargetProvider)
+	if err != nil {
+		return SessionResult{}, err
+	}
+	canonicalID, err := statePathSegment("target id", req.Bootstrap.Manifest.TargetID)
+	if err != nil {
+		return SessionResult{}, err
+	}
+	troot := bootTargetRoot
+	if filepath.Base(troot) != canonicalID ||
+		filepath.Base(filepath.Dir(troot)) != provider ||
+		filepath.Base(filepath.Dir(filepath.Dir(troot))) != t.Contract.TargetKind ||
+		filepath.Base(filepath.Dir(filepath.Dir(filepath.Dir(troot)))) != "targets" {
+		return SessionResult{}, fmt.Errorf("bootstrap target root %q is not the canonical targets/%s/%s/%s layout", bootTargetRoot, t.Contract.TargetKind, provider, canonicalID)
+	}
+	// The audit log must be the EXACT constructed path <TargetRoot>/
+	// workcell.audit.log, not merely under the target root: pointing it at another
+	// target-managed file (e.g. the record) would corrupt that file.
+	expectedAuditLog := filepath.Join(bootTargetRoot, "workcell.audit.log")
+	if filepath.Clean(req.Bootstrap.AuditLogPath) != expectedAuditLog {
+		return SessionResult{}, fmt.Errorf("bootstrap audit log %q is not the expected %q", req.Bootstrap.AuditLogPath, expectedAuditLog)
+	}
+	// The materialized workspace must be the exact constructed path and the manifest
+	// must agree, so a reconstructed Result cannot keep valid paths while pointing
+	// the session/audit at an unrelated real directory.
+	expectedWorkspace := filepath.Join(bootTargetRoot, "materializations", req.Materialization.Manifest.MaterializationID, t.Contract.WorkspaceMaterialization.WorkspaceDir)
+	if filepath.Clean(req.Materialization.MaterializedWorkspace) != expectedWorkspace {
+		return SessionResult{}, fmt.Errorf("materialized workspace %q is not the expected %q", req.Materialization.MaterializedWorkspace, expectedWorkspace)
+	}
+	if filepath.Clean(req.Materialization.Manifest.MaterializedWorkspace) != expectedWorkspace {
+		return SessionResult{}, fmt.Errorf("manifest materialized_workspace %q is not the expected %q", req.Materialization.Manifest.MaterializedWorkspace, expectedWorkspace)
+	}
+	// Stat the DERIVED canonical workspace (not the caller's spelling) through the
+	// openat-verified parent (statPathSafe: each parent O_NOFOLLOW, leaf Fstatat
+	// AT_SYMLINK_NOFOLLOW) rather than a fresh path-based Lstat, so a
+	// materializations/<id> parent swapped for a symlink after the earlier chain
+	// check cannot redirect the check; require a directory.
+	if st, err := statPathSafe(matStateRoot, expectedWorkspace); err != nil {
+		return SessionResult{}, err
+	} else if st.Mode&unix.S_IFMT != unix.S_IFDIR {
+		return SessionResult{}, fmt.Errorf("materialized workspace %q is not a real directory (symlinks are rejected)", expectedWorkspace)
+	}
+	recordPath := filepath.Join(bootTargetRoot, "sessions", sessionID+".json")
+	stateRoot := stateRootFor(bootTargetRoot)
+	// Reject a symlink on any target-managed dir below the state root, so a
+	// renamed-aside + symlinked parent cannot redirect the record write.
+	if err := rejectSymlinkChain(stateRoot, filepath.Dir(recordPath)); err != nil {
+		return SessionResult{}, err
+	}
+	// Serialize concurrent calls for this session id under an exclusive advisory
+	// lock so the read-then-write exactly-once guard is atomic.
+	unlock, err := lockSession(stateRoot, recordPath)
+	if err != nil {
+		return SessionResult{}, err
+	}
+	defer unlock()
+	targetID := req.Bootstrap.Manifest.TargetID
+	ws := expectedWorkspace // the derived canonical workspace, not the caller's spelling
+	// Recovery-aware idempotency: the record write and the audit append are two steps
+	// under this flock, so a crash between them can leave a record with no start
+	// events — which would dead-end a retry ("already started" here, then "missing
+	// start events" in FinishSession). If the record already exists AND belongs to
+	// this session/target, heal-or-return idempotently instead of hard-rejecting.
+	if _, statErr := os.Lstat(recordPath); statErr == nil {
+		current, err := readSessionRecordSafe(stateRoot, recordPath)
+		if err != nil {
+			return SessionResult{}, err
+		}
+		// Only recover a record that equals what THIS request would write, on every
+		// semantically-significant field the writer sets (identity + all contract
+		// fields + workspace/origin + audit log + assurance + control plane), so a
+		// retry with a DIFFERENT request (foreign identity, other materialization,
+		// changed assurance, etc.) is a distinct start and is rejected — not recovered.
+		// Only the re-stampable timestamps (started_at/observed_at) are excluded.
+		if !t.startedRecordMatchesRequest(current, req, sessionID, targetID, ws, expectedAuditLog) {
+			return SessionResult{}, fmt.Errorf("session %q already started with a divergent record (different request, not a retry)", sessionID)
+		}
+		// Are the start events already present? Read via the hardened reader; a missing
+		// audit log means they are absent (partial start). auditHasEvents parses exact
+		// event= tokens, so a torn final crash line is tolerated.
+		var sessionLines []string
+		if auditData, rerr := readAuditLog(stateRoot, expectedAuditLog); rerr == nil {
+			sessionLines = filterAuditSessionLines(auditData, sessionID)
+		} else if !errors.Is(rerr, os.ErrNotExist) {
+			return SessionResult{}, rerr
+		}
+		// Presence of each start event is judged by its EXACT complete expected line, not
+		// just the event= token: a line truncated after `event=<name>` (a short/failed/
+		// ENOSPC append, or tampering) carries the token but is not the real event, so
+		// token-only matching would wrongly treat it as satisfied and skip it. Matching
+		// the whole expected line makes torn/partial start lines unrecognized → absent →
+		// re-appended complete. The lines are stamped with the PERSISTED started_at (the
+		// first attempt's time) so the recovered audit stays consistent with the committed
+		// record; it is attacker-influenceable on-disk state, so validate it before use.
+		if err := validateAuditToken("persisted started at", current.StartedAt); err != nil {
+			return SessionResult{}, fmt.Errorf("session %q record holds an invalid started_at, refusing to recover: %w", sessionID, err)
+		}
+		// Reconcile: reject a conflicting complete line and append only the ABSENT exact
+		// lines (idempotent when all present). Stamped with the PERSISTED started_at so the
+		// recovered audit matches the committed record.
+		expected := t.startEventLines(current.StartedAt, sessionID, targetID, ws, req)
+		if _, err := t.reconcileStartEvents(stateRoot, expectedAuditLog, sessionID, expected, sessionLines); err != nil {
+			return SessionResult{}, err
+		}
+		return SessionResult{Record: current, RecordPath: recordPath, AuditLogPath: expectedAuditLog}, nil
+	} else if !os.IsNotExist(statErr) {
+		return SessionResult{}, statErr
+	}
+	// Fresh start (no record yet). Read the shared log (mode-healed if write-only); it may
+	// already hold start lines for this session id — pre-planted, or the record was deleted
+	// while the append-only log kept the events. Create the record via the SAME field map
+	// startedRecordMatchesRequest compares against, then reconcile: reject a divergent
+	// complete line and append ONLY the ABSENT exact lines, never duplicating a pre-existing
+	// exact one. On an audit-append failure the record is left in place (not removed) so the
+	// retry heals via the recovery path above rather than fresh-creating a duplicate.
+	var existing []string
+	if data, rerr := readAuditLog(stateRoot, expectedAuditLog); rerr == nil {
+		existing = filterAuditSessionLines(data, sessionID)
+	} else if !errors.Is(rerr, os.ErrNotExist) {
+		// Only genuine ABSENCE means "no evidence, proceed". Any other read error means the
+		// log EXISTS but is unusable — a symlink/FIFO/hard-link the hardened reader rejects,
+		// an un-healable EACCES, etc. Publishing a running record over an unusable/hostile
+		// audit log is wrong; refuse before creating the record.
+		return SessionResult{}, fmt.Errorf("session %q audit log at %s is present but unreadable — refusing to start: %w", sessionID, expectedAuditLog, rerr)
+	}
+	// Validate the pre-existing audit evidence BEFORE publishing the record, so a fresh
+	// start over bad evidence never leaves a published running record behind. Two checks,
+	// fail-closed and pre-publish: (a) the log already records this session as FINISHED (its
+	// record was lost/deleted after it completed) — refuse rather than resurrect it; and
+	// (b) a conflicting COMPLETE start line for any event — refuse rather than publish over
+	// divergent evidence. Only once the evidence is confirmed usable do we create the record
+	// and append the missing start events.
+	expected := t.startEventLines(req.StartedAt, sessionID, targetID, ws, req)
+	if auditHasEvents(existing, "session_finished") {
+		return SessionResult{}, fmt.Errorf("session %q already terminated; refusing to fresh-start over terminal evidence", sessionID)
+	}
+	for _, e := range expected {
+		if conflictingCompleteLine(existing, e.name, e.line) {
+			return SessionResult{}, fmt.Errorf("session %q has conflicting complete start evidence for %s — refusing to start", sessionID, e.name)
+		}
+	}
+	if err := writeSessionRecordAtomic(stateRoot, recordPath, t.startedRecordFields(req, sessionID, targetID, ws, expectedAuditLog), true); err != nil {
+		return SessionResult{}, err
+	}
+	if _, err := t.reconcileStartEvents(stateRoot, expectedAuditLog, sessionID, expected, existing); err != nil {
+		return SessionResult{}, err
+	}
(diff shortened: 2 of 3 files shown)


omkhar:
@codex review

omkhar:
Fixed in c63a075: StartSession now also requires the materialization and bootstrap to share the same target ROOT (not just target_id). The record and audit log are written under Bootstrap.TargetRoot, so a materialization produced under a different StateRoot would export lifecycle evidence split across two state trees — now rejected before any write. The conformance harness always uses one StateRoot for both phases (RunConformance passes c.StateRoot to Materialize and Bootstrap), so this is transparent there. TestStartSessionRejectsMismatchedTargetIDs now covers mismatched id, mismatched root (same id), and the matching-id-and-root happy path; corrected TestSessionTokenRejection to share one root so it still exercises token rejection. Green + shape 411.

omkhar:
@codex review

omkhar:
Addressed the P2 (even though the LGTM already matched head): StartSession now verifies the materialized workspace still exists as a directory before recording the session + workspace_materialized event, so a workspace removed/replaced between MaterializeWorkspace and StartSession is rejected rather than exporting evidence for a vanished path. TestStartSessionRejectsMissingWorkspace removes the workspace post-materialization and asserts StartSession errors. Green + shape.

omkhar:
@codex review

omkhar:
Comprehensive idempotency/retry hardening in 5294526: (1) workspace check now uses os.Lstat + IsDir so a symlink-swapped workspace is rejected (must be the real copied tree); (2) StartSession rejects an already-existing session record — exactly-once creation, no duplicate start triplet on retry; (3) FinishSession rejects a session already in FinalStatus — exactly-once finalization, no duplicate session_finished line. Proactive audit also caught a sibling: appendAuditLine now Lstats the log path and rejects a symlink before opening, so a swapped log can't redirect evidence to an attacker-chosen file. Manifest-stat TOCTOU audited (content comes from in-memory Result structs, existence-only stat → no effect). Tests: symlinked-workspace reject, full idempotency (2nd start/finish error, no dup events), symlinked-log reject — each proven by neutralization. Green + shape 553.

omkhar:
@codex review

omkhar:
Resolved all 5 at the design level in a97c362 — the exactly-once guards + audit-log symlink reject had created a partial-failure trap (record committed before audit → append failure blocked the retry forever). Fix makes the record commit atomic with its audit via rollback (guards intact): StartSession writes the 3 start audit lines as one append and, on failure, removes the record it just created (safe — the pre-existing-record guard proves this call created it); FinishSession captures the pre-finish record and restores it if the finish append fails. New rejectSymlinkDir Lstats $TargetRoot/sessions before each write. Invariant: after any call (success or failure), record present ⇔ its audit lines present, so a retry after a transient audit failure always progresses and never duplicates. Residual double-fault atomicity (rollback itself failing) needs journaling — out-of-scope for this deterministic single-shot target (the harness/probe never hit it), documented as such. Tests: start/finish rollback-on-audit-failure + retry-succeeds, symlinked-sessions-dir reject, idempotency preserved — proven by neutralization. Green + shape 687.

omkhar:
@codex review

omkhar:
Addressed: FinishSession now appends the session_finished event to the PERSISTED record's audit_log_path (its authoritative source of truth, already read for the idempotency check) instead of the caller-supplied req.Started.AuditLogPath — so a stale/reconstructed SessionResult can't split the finish event to a different or recreated log; start and finish evidence always share one log. Errors clearly if the persisted record somehow has no audit log path. TestFinishSessionUsesPersistedAuditLog passes a wrong AuditLogPath in the started result and asserts the finish line lands in the persisted log and nothing is written to the wrong path. Green + shape.

omkhar:
@codex review

omkhar:
**Disposition — converging (these 4 are out of scope for this deterministic evaluation target).** All prior valid findings across this PR's review are fixed (cross-target id+root matching, exactly-once start/finish guards, record⇔audit atomicity via rollback, workspace/log/sessions-dir symlink rejection, persisted-audit-log source-of-truth). These 4 remaining findings all describe **concurrent or adversarial-mid-lifecycle** scenarios that do not arise for this target:

- **Concurrent StartSession/FinishSession (L104, L183)** — the `AppleContainerTarget` is a deterministic reference implementation driven strictly **single-shot and sequentially** by `RunConformance` (one materialize→bootstrap→start→finish per session) and by the live probe. Overlapping same-`session_id` calls never occur. Making them race-safe needs OS file locking / `O_EXCL` exclusive creation in the shared `sessions` writer — significant machinery beyond an evaluation reference.
- **Audit log deleted between start and finish (L209)** — requires an external mid-lifecycle deletion the harness/probe never perform. Moreover the conformance validator (`validateAuditEvents`) requires the **exact required event sequence in order**, so a finish that dropped the start triplet would *fail conformance* — the evidence-integrity concern is already caught.
- **Parent-chain symlink swap after bootstrap (L95)** — an adversarial rename+symlink of a parent dir mid-lifecycle; full defense is a realpath check of the entire parent chain on every write, again production-backend hardening.

C1 ships this backend **preview-only / launch-blocked** (support-invisible); it is an evaluation, not a supported concurrent runtime. Per the review-loop round cap (this is well past 5 rounds, all findings P2, no P0/P1), recording this decision and merging on the resolved single-shot correctness. If Apple `container` is ever promoted past preview (the deferred B6 certification track), concurrent/adversarial hardening of the target is in scope then.

omkhar:
**Superseding the disposition above — addressing these properly.** On reflection (and direction), I'll harden the target for the concurrent + adversarial-mid-lifecycle cases rather than scope them out: per-session file-lock serialization so overlapping StartSession/FinishSession for the same session_id can't race (making the exactly-once guards atomic), a pre-finalize check that the audit log still exists with its start events, and a full parent-chain symlink check on the target root. Fixes incoming.

omkhar:
Hardened all 4 in 38e0c03 (target-local; sessions package untouched; -race green incl. concurrency tests at -count=20): (L104/L183) per-session advisory file lock — lockSession takes unix.Flock(LOCK_EX) on <sessionsDir>/<sessionID>.lock around the entire read-guard-then-write critical section of StartSession AND FinishSession, so overlapping same-session_id calls serialize and the exactly-once guards become atomic (second caller sees committed state → rejected); the .lock file is ignored by the .json-only session lister so conformance is unaffected. (L209) FinishSession now requires the persisted audit log to still carry this session's workspace_materialized+bootstrap_ready+session_started events before finalizing — a deleted/truncated log errors instead of exporting a finish-only record. (L95) rejectSymlinkChain Lstats every target-managed dir from the state root down to sessions (stopping at the state root so legit system symlinks like /var above it don't false-positive). Tests: concurrent-exactly-once (both start and finish), deleted-log-rejected, symlinked-parent-rejected — each proven by neutralization; single-shot happy path green. Shape 952.

omkhar:
@codex review

omkhar:
Fixed all 3 in a3f91ca (target-local, -race green): (L150) StartSession now revalidates the manifest-derived audit tokens (materialization_id/target_id/bootstrap_id/image_ref) from the in-memory Result via validateAuditToken before any write, so a reconstructed/stale Result with whitespace/newlines can't inject audit content; (L136) StartSession requires Bootstrap.AuditLogPath to be pathWithin the resolved Bootstrap.TargetRoot, so the start log (and exported audit_log_path) can't be written outside the state tree — mirrors the FinishSession source-of-truth fix; (L243) the rollback test now replaces the log with a symlink-to-sidecar so appendAuditLine's symlink-reject fails the append for ANY uid (root-safe) while SessionAuditRecords still finds the start triplet, exercising the rollback branch rather than the missing-log branch. Tests proven by neutralization; conformance single-shot + live probe green. Shape 1043.

omkhar:
@codex review

omkhar:
Fixed both in 71e850d (target-local, -race green): (L118) StartSession now recomputes the exact target-constructed workspace path (TargetRoot/materializations/<materialization_id>/<workspace_dir>) and requires req.Materialization.MaterializedWorkspace AND the manifest's own MaterializedWorkspace to equal it — so a reconstructed Result pointing the workspace at unrelated contents is rejected, even a self-consistent one (caught by the expected-path check); Lstat+IsDir retained. (L366) auditHasEvents now parses each line's exact event= field (new auditLineEvent helper, same approach as conformance's validateAuditEvents) instead of substring, so a field value ending in event=session_started can't forge the start-events check. Tests: foreign-workspace rejected (both result-only and result+manifest), forged-start-event rejected — proven by neutralization; conformance + probe green. Shape 1128.

omkhar:
@codex review

omkhar:
Fixed in ed3fde6 (target-local, -race green): replaced the too-loose 'AuditLogPath under TargetRoot' containment check with an EXACT-path match — recompute the bootstrap log path exactly as BootstrapTarget builds it (filepath.Join(Bootstrap.TargetRoot, "workcell.audit.log")) and require req.Bootstrap.AuditLogPath to equal it. This blocks a reconstructed Result from pointing AuditLogPath at another target-managed file (e.g. <TargetRoot>/sessions/<session>.json, which would pass containment and be corrupted by the audit append). Mirrors the MaterializedWorkspace exact-path fix; also simplifies (drops the resolveExistingPrefix/pathWithin dance). Test: AuditLogPath=record-path → StartSession errors and the record is never corrupted; proven by neutralization. Green + shape 1150.

omkhar:
@codex review

omkhar:
Updated to 58bec0a (rebased onto the now-merged audit-helpers PR #436; stack is now core→audit→lifecycle→conformance→eval). This carries the full lifecycle hardening accumulated during review: cross-target id+root matching, exactly-once StartSession/FinishSession under per-session flock, record⇔audit atomicity via rollback, and the input-trust closure (every caller-supplied path AND the on-disk manifests verified against the target-constructed state — the manifest byte-comparison that resolves the prior open finding). The audit-line helpers moved to #436 (merged); this PR is the 2 lifecycle files (target_session.go + test), 1066 lines. Green + -race clean.

omkhar:
@codex review

omkhar:
Closed the last input-trust gap in 258fa08: StartSession now verifies TargetRoot is the CANONICAL <stateRoot>/targets/<kind>/<provider>/<target_id> layout (trailing 4 components must equal manifest target_id / contract provider / contract kind / 'targets', with target_id+provider validated as single segments). Previously the checks only proved Materialization and Bootstrap AGREED on TargetRoot — a self-consistent reconstructed request could point it at an arbitrary dir, and stateRootFor + all manifest/record/audit paths derive from it. TestStartSessionRejectsNonCanonicalTargetRoot relocates the whole target subtree to a non-canonical leaf and rewrites every self-consistent path field + the on-disk manifest's materialized_workspace, so every OTHER guard still passes and only the canonical check catches it (proven by neutralization — confirms verifyPersistedManifest alone didn't close it). -race green + shape 1143.

omkhar:
@codex review

omkhar:
Closed both in 8c67a54 (-race green, FIFO test terminates): (L159) StartSession now applies the shared rejectSymlinkChain to the materialization parent chain (<targetRoot>/materializations/<id>) BEFORE reading the manifest, so a symlinked <id> dir → attacker tree can't defeat verifyPersistedManifest + the workspace check. (L285) FinishSession no longer reads the audit log via os.ReadFile (which HANGS on a FIFO); new readAuditLog (added to audit.go) does the atomic openat traversal + O_NOFOLLOW|O_NONBLOCK open + Fstat regular-file/Nlink==1, so a FIFO/symlink/hard-link/special log errors promptly and leaves the session non-final/retryable. This makes the audit READ path as hardened as the WRITE path. Tests: symlinked-materialization-dir rejected; FIFO-log FinishSession errors under a 5s timeout (no hang) — proven by neutralization. Shape 1200 (3 files incl. the audit.go read helper).

omkhar:
@codex review

omkhar:
@codex review

omkhar:
@codex review

omkhar:
Both closed in 161190c (-race green across sessions/remotevm/applecontainer): (L105) bootstrap manifest dir now gets rejectSymlinkChain before its read, symmetric with materialization. (L186) session RECORD write is now atomic: extracted an ADDITIVE sessions.EncodeSessionRecord (read-merge-normalize-validate-marshal — WriteSessionRecord's signature+behavior unchanged, so remotevm is byte-for-byte identical and green), and applecontainer's writeSessionRecordAtomic writes those validated bytes via the openat traversal (O_CREAT|O_EXCL|O_NOFOLLOW for StartSession create-once; O_WRONLY|O_NOFOLLOW|O_NONBLOCK + Fstat regular/Nlink==1 + truncate for FinishSession rewrite/rollback). Every I/O site in the target — record + audit reads and writes, workspace, all manifest dirs — now goes through verified-fd no-follow/regular-file guards. Record-atomic tests on the fstest branch. shape 1090.

omkhar:
@codex review

omkhar:
Both closed in 0c6a3bd (-race green): (L268) session record READS are now hardened like the audit read — additive sessions.DecodeSessionRecord extraction (ReadSessionRecord unchanged, remotevm unaffected) + readFileSafe (openat O_RDONLY|O_NOFOLLOW|O_NONBLOCK + Fstat regular/Nlink==1); all four record reads (StartSession re-read, FinishSession pre-finalize + rollback capture + final re-read) go through it, so a FIFO record can't hang and a symlinked/hardlinked record is refused. (L293) record REWRITE is now atomic stage-and-rename (unix.Renameat over the record from a verified parent fd, unique O_EXCL temp) instead of truncate-in-place — a mid-write failure leaves the original intact; existing-record symlink/FIFO/hardlink still refused. Record and audit I/O are now fully symmetric: every read FIFO/symlink-safe, every write atomic + no-follow. Tests on fstest (inode-replacement proves stage-and-rename); proven by neutralization. shape 1146.

omkhar:
@codex review

omkhar:
Closed the last unhardened read in 6649182 (-race green): the record-write path's INTERNAL read-existing (inside EncodeSessionRecord's merge) was still os.ReadFile. Added additive sessions.EncodeSessionRecordFrom(existingBytes, updates) (EncodeSessionRecord/WriteSessionRecord/ReadSessionRecord signatures unchanged → remotevm byte-identical); writeSessionRecordAtomic now reads the existing record via the hardened readFileSafe (openat/O_NOFOLLOW/O_NONBLOCK/Fstat regular/Nlink==1) then EncodeSessionRecordFrom. No read anywhere on the write path uses unhardened os.ReadFile. This completes it: EVERY read (standalone + read-existing-within-write) and EVERY write (atomic openat create-once / stage+renameat) in the target is FIFO/symlink/hardlink-safe. Tests: FIFO/symlinked existing-record on rewrite → prompt error, no hang/follow (proven by neutralization). shape 1185.

omkhar:
@codex review

omkhar:
Comprehensive filesystem-op sweep in 7a95e7d (-race green): fixed the 3 reported (manifest FILE reads → readFileSafe; createRecordAtomic unlinks on write failure; StartSession rollback → removeRecordAtomic openat Unlinkat) AND audited EVERY path-based op on a target-managed path. Result: no unhardened path-based op remains in the session-lifecycle path — all reads via readFileSafe (openat/O_NOFOLLOW/O_NONBLOCK/Fstat regular/Nlink==1), all writes atomic (O_EXCL create-once + unlink-on-fail / stage+renameat), lock file openat-safe, rollbacks openat-safe. Bare Lstats are fast-error paths whose real op is openat-guaranteed. The Materialize/Bootstrap CREATION-phase writes run before StartSession, which then re-validates the whole tree (rejectSymlinkChain full chain + canonical-layout + hardened manifest reads that reject a symlinked manifest + workspace Lstat), so it refuses to start on a tampered tree. Source-workspace reads are outside stateRoot (by design). Tests on fstest, each proven by neutralization. shape 1182.

omkhar:
@codex review

omkhar:
Persisted-field-trust class swept in 89ee450 (-race green): FinishSession no longer trusts mutable fields from the persisted record. audit_log_path is RE-DERIVED as filepath.Join(targetRoot, 'workcell.audit.log') for both the start-events read and the finish append (a tampered field can't redirect/DoS the finish). session_id and target_id are re-derived (from the record filename / canonical targetRoot base) AND asserted-equal to the persisted values (mismatch → reject). status is legitimately-stored idempotency state (kept). Full per-field disposition confirmed no other persisted field locates a file or builds identity. StartSession already asserts AuditLogPath canonical before any write. Tests on fstest: tampered audit_log_path → finish lands in canonical log, decoy untouched; tampered session_id/target_id → rejected; all proven by neutralization. shape 1195.

omkhar:
@codex review

omkhar:
Fixed in dd7e449 (-race green): FinishSession's exactly-once guard now rejects ANY terminal status (exited/failed/aborted) via the additive-exported sessions.IsTerminalSessionStatus, not just the contract FinalStatus — so a record already terminal in a non-exited state can't be re-finalized. Test on fstest tampers a fully-valid terminal 'failed' record and asserts rejection (strengthened after neutralization showed a status-only tamper was rejected earlier at the record-read validation, masking the guard). Happy path running→exited still finalizes.

omkhar:
@codex review

omkhar:
Updated to 91c8c44 (rebased onto the now-merged io primitives PR #437; stack: io✅ → lifecycle → fstest → conformance → eval). This is now just StartSession/FinishSession/lockSession/verifyPersistedManifest (2 files, 950 lines) calling the merged hardened I/O primitives. Includes the 2 fixes from the prior round: workspace type-check via the verified-fd fstatat (no TOCTOU on a symlinked materialization parent), and persisted manifest contract-identity fields (target_kind/provider/transport/…) asserted against t.Contract (not just self-consistency). Green + -race clean.

omkhar:
@codex review

omkhar:
Fixed in 1b7b218 (-race green): FinishSession now HEALS the persisted record's audit_log_path — the finalize update writes the canonical filepath.Join(targetRoot, 'workcell.audit.log'), so a value tampered between start and finish no longer survives into the exported finalized record (both the finish event AND the record are now canonical). Test TestFinishSessionHealsTamperedAuditLogPath tampers the field to a decoy and asserts the finalized record is canonical — proven by neutralization. Rebased onto the now-merged io PR #437.

omkhar:
@codex review

omkhar:
Fixed in 9d98e23 (-race green): FinishSession now asserts the persisted record's target_kind == contract.TargetKind AND target_provider == contract.TargetProvider before finalizing, so the apple-container target refuses to finalize a record from another target (e.g. remote_vm/fake-remote) — the FinishSession record-side equivalent of StartSession's assertManifestContract. Test TestFinishSessionRejectsForeignTarget (foreign target_kind/provider → rejected), proven by neutralization.

omkhar:
@codex review

omkhar:
Comprehensive FinishSession input-trust parity in 4c75167 (-race green), bringing the finish side to full parity with StartSession: (L271) FinishSession now asserts its derived targetRoot is canonical (<stateRoot>/targets/<kind>/<provider>/<id>) before any read/write — a stale/forged RecordPath is rejected up front; (L300) the path-derived session_id/target_id are validateAuditToken'd (no whitespace) before entering the finish audit line; (L419) assertManifestContract now pins excluded_paths (==[".git"]) + version on the workspace manifest and version on bootstrap (kind/provider/assurance/support/runtime/access were already pinned). Included a StartSession↔FinishSession parity table in the report; every path/identity/audit-bearing field FinishSession reads is now re-derived or asserted. Tests: non-canonical-RecordPath, whitespace-session-id, tampered-excluded_paths — each proven by neutralization.

omkhar:
@codex review

omkhar:
Fixed in a374538 (-race green): root cause was a RAW STRING equality of Materialization.TargetRoot vs Bootstrap.TargetRoot — a canonical path with a trailing slash on one Result (same path) was wrongly rejected as a mismatch. Now both are filepath.Clean'd up front and the cleaned values are used consistently (14 usages: the equality, the canonical-layout assertion, stateRootFor, recordPath, both manifest paths, audit-log/workspace pins) plus FinishSession's targetRoot. Clean collapses trailing-slash/.///../, and it's not a bypass — the layout assertion still requires the exact trailing components + valid statePathSegment leaf, so a .. that changed components is still rejected. Tests: trailing-slash accepted (neutralize→raw compare→FAIL), ..-bearing still rejected.

omkhar:
@codex review

omkhar:
Comprehensive path-canonicalization sweep in 807775f (-race green): no caller-supplied path is trusted by spelling anymore. (L280) FinishSession Cleans Started.RecordPath and pins it EXACTLY to <targetRoot>/sessions/<session_id>.json — a record path under the same target root but outside sessions/ (audit log, a manifest, a sibling) is rejected. (L167) Bootstrap.AuditLogPath is Clean'd + pinned to <targetRoot>/workcell.audit.log, and the DERIVED value (not the caller spelling) is used for the append/persisted field/return. Also swept: both manifest paths, MaterializedWorkspace (result + manifest field), all Clean+equality-pinned to canonical derivations (full path-field table in the stack). Tests: record-path-outside-sessions rejected, /.-spelled audit-log + manifest-workspace accepted-as-canonical — each proven by neutralization.

omkhar:
@codex review

omkhar:
Fixed in d610445 (-race green): FinishSession now validates the path-derived session_id AND target_id with statePathSegment (a strict SUPERSET of validateAuditToken — it also rejects empty, '.', '..', absolute, and any '/' or '\'), so a degenerate record like sessions/..json (session_id '.') or a '\'-bearing basename is rejected with a clear segment error before any read/write/audit use, rather than only tripping the downstream RecordPath pin. Test TestFinishSessionRejectsNonSegmentSessionID (table: '.' and '\'), proven by neutralization (validateAuditToken-only → passes id check → FAIL).

omkhar:
@codex review

omkhar:
Fixed in a00d4c9 (-race green, concurrency test 30× non-flaky): StartSession is now recovery-aware idempotent under the per-session flock. On record-exists: (1) identity gate — session_id/target_id + contract kind/provider must match (foreign/tampered record still rejected, never recovered); (2) presence check via auditHasEvents on the hardened audit read (missing log = absent; torn crash line tolerated); (3) events present → idempotent success (no duplicate triplet); events absent (crash between record-write and audit-append) → re-append the triplet so the session is finishable; record absent → normal create. Also fixed a PRE-EXISTING lockSession ENOENT race (bounded retry) that was making the old concurrency test spuriously green — verified empirically on the unchanged commit. Tests: recover-partial-start, idempotent-when-fully-started, foreign-record-not-recovered — proven by neutralization.

omkhar:
@codex review

omkhar:
Comprehensive recovery-consistency pass in 4bbb19d (-race green, finish concurrency 20×): (L410) FinishSession is now symmetric recovery-aware idempotent — on an already-terminal record it verifies finish content (exit_status matches request; finished_at re-stampable), then if session_finished audit is absent (crash between terminal rewrite and audit append) re-appends it, else idempotent no-op; identity+terminal assertions moved before the check. (L227) StartSession recovery now requires the persisted record to equal what THIS request would write on ALL writer-set fields (session_id/profile/target_id/kind/provider/assurance/runtime/transport/agent/mode/status/workspace/origin/root/worktree/audit_log_path/initial+current_assurance/control_plane), excluding only re-stampable timestamps (started_at/observed_at) — a divergent request is rejected, not recovered. Also SPLIT the fstest suite into two reviewable test PRs (fstest-fs: filesystem hardening; fstest-lifecycle: input-trust+recovery). Tests: recover-partial-finish, idempotent-finish, divergent-exit-rejected, divergent-start-rejected, genuine-timestamp-only-retry-recovers — proven by neutralization.

omkhar:
@codex review

omkhar:
Fixed in 151a873 (-race green): FinishSession recovery no longer fabricates an audit log when evidence was deleted. It now only heals a GENUINE crash-partial (log intact with the start events present but missing only session_finished → re-append). If the audit log is missing (os.ErrNotExist)/non-regular OR the start events are absent → REFUSE with a clear evidence-missing error; never create a provenance-less finish-only log. Documented the correct asymmetry in-code: StartSession may (re)create a missing log (the start triplet is the FIRST write), but FinishSession requires the start events to pre-exist. Tests: refuse-when-log-deleted (creates no log), refuse-when-start-events-gone, still-recovers-genuine-partial, idempotent-when-finished — proven by neutralization.

omkhar:
@codex review

omkhar:
Fixed both in bcd52c0 (-race green): (L244) recovery now re-appends the start triplet with the PERSISTED started_at (current.StartedAt), not the retry's time — factored startTriplet(ts,...) so the fresh path uses req.StartedAt and the crash-partial recovery path uses the persisted first-attempt ts, keeping the healed audit line consistent with the committed record. (L229) idempotency now matches BootstrapID + ImageRef: since those are audit-line-only (not in the record), added auditLineFieldValue() (same exact-token parsing as auditHasEvents) to read them from the persisted bootstrap_ready line, and the fully-started idempotent branch rejects a retry whose bootstrap_id/image_ref differ (different bootstrap = different start); the recovery/first-write branch is unaffected. Tests: recovery-uses-persisted-ts, rejects-divergent-bootstrap-on-retry — proven by neutralization.

omkhar:
@codex review

omkhar:
Fixed in 9d51685 (-race green): the FinishSession idempotent/recovery branch now requires current.Status == t.Contract.Session.FinalStatus (the exact status this finish writes, 'exited') — a record terminal in a DIFFERENT status (failed/aborted) is rejected ('already finalized with a different terminal status'), not treated as idempotent-success or recovered. IsTerminalSessionStatus still gates 'is it terminal at all' (vs a non-terminal record that finalizes); this narrows only the idempotent-success path to the matching FinalStatus. Test strengthened so the tampered failed record has a matching exit_status → only the status gate rejects it — proven by neutralization.

omkhar:
@codex review

omkhar:
Updated to 22e4f97 (rebased onto the now-merged recovery PR #439; stack: recovery✅ → lifecycle → fstest-fs → fstest-lifecycle → conformance → eval). Lifecycle is now just the StartSession/FinishSession orchestration + verifyPersistedManifest/assertManifestContract (1045 lines) calling the merged recovery/idempotency helpers; it carries all the accumulated fixes from the recovery rounds (canonical-path pins, full-record exhaustive match, symmetric crash recovery, evidence-integrity, finish-status gate). Green + -race clean.

omkhar:
@codex review

omkhar:
Updated to acf9bd8 (rebased onto the now-merged record PR #440; stack: record✅ → lifecycle → fstest-fs → fstest-lifecycle → conformance → eval). Lifecycle is the StartSession/FinishSession orchestration (target_session.go + test, 1065 lines) consuming the merged bootstrap_id/image_ref record fields — bootstrap-drift on a recovery/idempotent retry is now caught by the exhaustive record match (even before any audit line lands), the L226 belt-and-suspenders bootstrap_ready check, and the L417 conflicting-finish-evidence rejection. Green + -race clean.

omkhar:
@codex review

omkhar:
Updated to de819f6 (-race green; rebased onto current main after record #440 merged — main is now a clean ancestor, lifecycle is the bottom PR). Fixes for the audit torn-write edges: (L204) recovery presence now requires the EXACT COMPLETE start line (slices.Contains on the full expected line), not just the event= token — a start line truncated after event=<name> is treated as absent and the complete line re-appended; the bootstrap belt-and-suspenders fires only on a COMPLETE bootstrap_ready line that differs (a torn fragment heals). (L258) the start triplet is a single 3-line buffer written in one io.WriteString (no partial-triplet window); we deliberately do NOT truncate the shared per-target log (racy vs the per-session flock) and rely on exact-line matching to ignore torn garbage. (L417/L477) the conflicting-finish guard rejects only a DIFFERENT COMPLETE session_finished line; a fragment torn after the token is treated as no-finish → healed, neither falsely idempotent nor rejected. Tests (proven by neutralization): rewrites-truncated-start-line, rejects-different-complete-finish, heals-torn-finish-fragment (+ after-rollback), single-buffer-triplet.

omkhar:
@codex review

omkhar:
Fixed both in b097db5 (-race green): the rollback-after-append-failure paths were forcing duplication on retry, so both now leave a recoverable state for the retry to heal. (L257) StartSession no longer removes the record on an audit-append failure — the durable record + whatever complete start lines landed are left as a crash-partial; the retry recovers via exact-line append-only-missing (0-lines and partial-lines handled uniformly), no duplicate events. (L477) FinishSession no longer rolls the record back to non-terminal on an audit-append failure — the terminal rewrite committed (the session IS finished), so it stays terminal and the retry's terminal-recovery heals the audit (complete finish line present → idempotent; absent → re-append one), eliminating the double-finish. Both still return the append error so the caller retries. Tests via an appendAudit failpoint seam (lands N complete lines then errors): keeps-partial-and-recovers, stays-terminal-idempotent, stays-terminal-recovers-when-no-line — proven by neutralization.

omkhar:
@codex review

omkhar:
Fixed in 9ae612c (-race green): both FinishSession start-provenance guards (fresh-finish L459 AND terminal-recovery L441) now require EXACT COMPLETE start lines instead of token matching, via a new shared startEventsComplete helper. It reconstructs the three expected complete start lines from the PERSISTED record using the same startEventLines renderer StartSession uses (no duplicated rendering; bootstrap_id/image_ref/workspace_origin from the persisted record, materialization id recovered as Base(Dir(workspace)) — safe since the contract pins workspace_dir to the single segment 'workspace'), so a start line truncated after event=session_started is no longer accepted → FinishSession refuses ('missing or incomplete start events'). Consistent with the StartSession-side L204 exact-line fix. Test TestFinishSessionRefusesTornStartLine, proven by neutralization.

omkhar:
@codex review

omkhar:
Fixed both in 0e370b5 (-race green): the idempotent-success early-return no longer bypasses conflict detection. Added a shared conflictingCompleteLine(lines, event, want) helper (completeness key derived from want's own trailing field, so it's generic across start events and the finish line; a torn fragment is not complete → ignored/healed). (L229) StartSession's fully-started idempotent path now scans for a second, DIFFERENT complete start line before returning — rejects 'conflicting complete start evidence' (this subsumes the old divergent-bootstrap belt-and-suspenders, which was removed). (L456) FinishSession runs the different-complete-finish-line check BEFORE the idempotent slices.Contains return. Tests: rejects-conflicting-complete-start, rejects-conflicting-complete-finish-on-idempotent-retry, idempotent-without-conflict-still-succeeds — proven by neutralization.

omkhar:
@codex review

omkhar:
Comprehensive close-out in 40548e6 (-race green). Includes a P1 fix surfaced+empirically-reproduced by an adversarial audit AND your L244/L424:
- (P1 / L244) appendAuditLine now heals the append boundary: before writing, if the log's tail byte isn't a newline (a prior short/ENOSPC append left a newline-less fragment), it prepends a separator so the new line lands on its own line instead of MERGING onto the fragment. The merge pathology (merged line never byte-matches the expected line yet inherits its trailing field → conflictingCompleteLine false-positive → PERMANENT session dead-end + silent evidence corruption) is closed. fd opened O_RDWR for the tail Pread; benign shared-log race (worst case an extra empty line, ignored). Repro test: two-append merge → clean line on its own line (neutralize → merge → FAIL).
- (P2) fresh StartSession + fresh non-terminal FinishSession now run the conflictingCompleteLine scan before appending (the guard was only on the recovery branches) — a pre-planted conflicting complete start/finish line is refused fail-closed instead of silently accepted then dead-ending later retries. Absent/unreadable log is skipped to preserve record-kept-for-recovery.
- (L424) FinishSession now asserts the FULL contract identity (assertRecordContract: kind/provider/assurance/runtime/transport/control-plane) on a record under the canonical path, not just kind/provider.
Also split the hardening test suite (fstest-lifecycle → -a input-trust/path + -b recovery/idempotency/conflict). All proven by neutralization; -race green incl remotevm/sessions/metadatautil.

omkhar:
@codex review

omkhar:
Updated to 8905314 (rebased onto the now-merged audit-newline PR #441 which carries the P1 append-boundary heal + flock + write-only-recovery + non-blocking-lock fixes; stack: audit-newline✅ → lifecycle → fstest-fs → fstest-lifecycle-a → fstest-lifecycle-b → conformance → eval). Lifecycle is the StartSession/FinishSession orchestration (target_session.go + test, 1173) carrying: fresh-path conflictingCompleteLine scans (StartSession+FinishSession, closing the fresh-branch asymmetry an adversarial audit found), assertRecordContract full contract-drift rejection (L424), and the L333 torn-fragment-PREFIX exclusion in conflictingCompleteLine (a truncated fragment is a prefix of the expected line → ignored/healed; only a diverging line conflicts). Green + -race clean.

omkhar:
@codex review

omkhar:
Updated to 1e38538 (rebased onto the now-merged audit-read-heal PR #442; stack: audit-read-heal✅ → lifecycle → fstest-fs → fstest-lifecycle-a → fstest-lifecycle-b → conformance → eval). Lifecycle is the StartSession/FinishSession orchestration (target_session.go + test, 1189) carrying the full accumulated hardening: fresh-path + recovery conflict scans (shared reconcileStartEvents, append-only-missing, no exact-duplicate), full contract-drift rejection (assertRecordContract), L333 torn-fragment-prefix tolerance in conflictingCompleteLine, persisted-render-field token validation in startEventsComplete, and (via the merged audit primitives) write-only-log read/write heal + flock serialization. Green + -race clean.

omkhar:
@codex review

omkhar:
Updated to 243e244 (-race green): (L246) fresh-start now REFUSES over terminal evidence — if the record was lost/deleted but the append-only log still has a session_finished event, StartSession refuses ('already terminated; refusing to fresh-start over terminal evidence') rather than resurrecting a completed session. Test TestFreshStartRefusesOverTerminalEvidence, proven by neutralization. (L353) pushing back — declining the sentinel fix (see the inline reply): the prefix-exclusion is a deliberate P1 torn-write-DoS-avoidance tradeoff; the complete-prefix-value edge is near-zero impact (record authoritative, correct line still appended, only defense-in-depth conflict-rejection lost for one narrow shape) and the only robust fix is a durable audit-format sentinel that couples to the conformance parser + remotevm's byte-identical format + hardcoded tests — disproportionate for a preview-only target.

omkhar:
@codex review

omkhar:
Updated to f7d20db (rebased onto the now-merged session-helpers PR #443, which extracted the render/compare/validate helpers AND landed the audit-line sentinel that resolves the L353/L118 prefix-value concern; stack: session-helpers✅ → lifecycle → fstest-fs → fstest-lifecycle-a → fstest-lifecycle-b → conformance → eval). Lifecycle is now the StartSession/FinishSession orchestration (target_session.go + test, 1077) carrying: the two finish-render sentinels, L551 (idempotent-finish checks start provenance before returning), L249 (fresh-start validates evidence before publishing the record — no orphan on bad evidence), L246 (fresh-start refuses over terminal evidence), and the full accumulated hardening via the merged helpers/primitives. The earlier L353 push-back is superseded — it's now fixed via the sentinel. Green + -race clean.

omkhar:
@codex review

omkhar:
Fixed in 16e6f33 (-race green): the fresh-start pre-publish read now distinguishes absence from an unusable log. errors.Is(rerr, os.ErrNotExist) → genuinely no evidence → proceed (publish record + append triplet); ANY OTHER read error (symlink/FIFO/hard-link rejection, un-healable EACCES, etc.) → the log is present-but-unusable → REFUSE before publishing ('present but unreadable — refusing to start'), no record created. FinishSession's reads already fail-closed on any read error (verified — no symmetric gap). Test TestStartSessionRefusesUnreadableLogBeforePublish (symlinked log → refused + no record; genuinely-absent → fresh start succeeds), proven by neutralization (treat any error as absent → record published → FAIL).

omkhar:
@codex review

omkhar:
Fixed all 3 in 7e25a08 (-race green): (L371) terminal-path now rejects a record whose final_assurance != contract assurance (targeted check alongside the terminal status/exit gates; assertRecordContract runs on non-terminal records with empty final_assurance so the check is on the terminal path). (L495) fresh (non-terminal) finish appends the freshFinish line only if not already present (slices.Contains) — no duplicate finish line, mirroring the start-path reconcile. (L464) startEventsComplete now runs conflictingCompleteLine per start event before the presence check, so a divergent complete session_started line alongside the expected triplet makes FinishSession refuse ('conflicting complete start evidence') — both call sites (terminal-recovery + fresh) inherit it. Tests: drifted-final-assurance-rejected, no-duplicate-finish-line, conflicting-start-evidence-rejected — proven by neutralization.

omkhar:
@codex review

PR #436, opened 2026-07-07
^F C1 (2/5): apple-container audit-line helpers

omkhar:
**Roadmap C1 — Apple `container` backend evaluation, part 2 of 5** (stack: contract+materialization/bootstrap → **audit helpers** → session lifecycle → conformance → probe/guard+evaluation). Part 1 merged in #434.

## What
- **audit.go** — the audit-line helpers used by the session lifecycle: `appendAuditLine` (whitespace-delimited `key=value` writer with symlink rejection), `encodeAuditPathValue`/`decodeAuditPathValue` (percent-encode path values so spaces/newlines can't forge tokens or lines while paths round-trip), and `auditHasEvents`/`auditLineEvent` (exact `event=` field parsing).
- Extracted from the lifecycle PR to keep each PR under the reviewability gate; the lifecycle methods that consume these land in part 3.

## Validation
`go build`/`vet`/`gofmt`, `go test ./internal/applecontainer/` (audit-path round-trip, exact-event match), `go test -race` — all green; 2 files / 180 lines.

### DIFF
--- a/internal/applecontainer/audit.go
+++ b/internal/applecontainer/audit.go
@@ -1,0 +1,201 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+package applecontainer
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"golang.org/x/sys/unix"
+)
+
+// encodeAuditPathValue makes a filesystem path safe to interpolate into a
+// whitespace-delimited audit line without rejecting legitimate spaces. It ranges
+// over BYTES, not runes, so a non-UTF-8 path byte is preserved exactly instead of
+// collapsing to U+FFFD: any '%', any byte <= 0x20 (space and all control/
+// whitespace), and any byte >= 0x7f (DEL and every high/non-UTF-8 byte) is
+// percent-encoded; other printable ASCII is emitted literally. A space cannot
+// inject a stray token and a newline cannot forge a line, while
+// decodeAuditPathValue recovers the exact original byte string for ANY input.
+func encodeAuditPathValue(value string) string {
+	var b strings.Builder
+	for i := 0; i < len(value); i++ {
+		c := value[i]
+		if c == '%' || c <= 0x20 || c >= 0x7f {
+			fmt.Fprintf(&b, "%%%02X", c)
+			continue
+		}
+		b.WriteByte(c)
+	}
+	return b.String()
+}
+
+func decodeAuditPathValue(value string) string {
+	var b strings.Builder
+	for i := 0; i < len(value); {
+		if value[i] == '%' && i+3 <= len(value) {
+			if n, err := strconv.ParseUint(value[i+1:i+3], 16, 8); err == nil {
+				b.WriteByte(byte(n))
+				i += 3
+				continue
+			}
+		}
+		b.WriteByte(value[i])
+		i++
+	}
+	return b.String()
+}
+
+// auditHasEvents reports whether every named event appears as the exact parsed
+// event= field of some record (lines already filtered to one session by
+// SessionAuditRecords), so a value like a path ending in "event=session_started"
+// cannot satisfy it — only a genuine event line counts.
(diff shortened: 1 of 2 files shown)


omkhar:
@codex review

omkhar:
Fixed the TOCTOU in appendAuditLine: added syscall.O_NOFOLLOW to the OpenFile flags so the open itself atomically refuses to follow a symlink on the final path component — closing the race where the log path is swapped to a symlink between the pre-open Lstat and the open. The Lstat is kept for a clear error on the common (non-racy) path; O_NOFOLLOW is the atomic backstop. macOS target, O_NOFOLLOW is supported. Existing TestAppendAuditLineRejectsSymlink (symlink at the path → rejected, no write-through) stays green under both guards.

omkhar:
@codex review

omkhar:
Closed the parent-dir vector in 61c3487: introduced a single shared rejectSymlinkChain(trustedRoot, dir) in audit.go that Lstats each component from dir up to (not including) trustedRoot — so a symlinked target-managed parent of the audit log is rejected, while the walk STOPS at the trusted StateRoot so the macOS /var→/private/var ancestor above it never false-positives. appendAuditLine now takes trustedRoot (derived via stateRootFor) and calls it before MkdirAll, keeping the final Lstat+O_NOFOLLOW as the atomic last-component guard. Consolidation bonus: the lifecycle's duplicate fixed-depth rejectSymlinkChain was deleted and now uses this shared one (lifecycle shrank 1088→1066). Test TestAppendAuditLineRejectsSymlinkedParent (symlinked parent rejected; /var-ancestor happy path not flagged) proven by neutralization; -race green.

omkhar:
@codex review

omkhar:
Terminal symlink fix in e2d9d26 (atomic openat; -race green): appendAuditLine now opens the trusted StateRoot by path (O_DIRECTORY, no O_NOFOLLOW so the legit macOS /var ancestor traverses), then openat's EACH target-managed component below it and the leaf log with O_NOFOLLOW|O_DIRECTORY|O_CLOEXEC relative to the verified parent fd — no path re-resolution, so no swap between check and use is possible (closes the residual TOCTOU that the Lstat-chain left). rejectSymlinkChain is kept for the lifecycle's session-record write. Also fixed encodeAuditPathValue to range over BYTES (percent-encoding %, <=0x20, >=0x7f) so non-UTF-8 POSIX filenames round-trip exactly instead of collapsing to U+FFFD. Tests: symlinked-parent openat-rejected + happy path; byte-preserving round-trip for non-UTF-8/high/DEL/whitespace — proven by neutralization. Shape 329.

omkhar:
@codex review

omkhar:
Closed the FIFO/device edge in 68bc9e6 (-race green, test terminates promptly): the leaf openat now sets O_NONBLOCK (so a reader-less FIFO returns ENXIO immediately instead of blocking O_WRONLY) and, on the returned fd, unix.Fstat's and requires S_IFREG — a reader-present FIFO/device/socket is rejected with 'not a regular file'. The check inspects the actual opened object (no path re-lookup → race-free). Test holds a FIFO read-end open + a 5s select timeout so a hang FAILS rather than passes; proven by neutralization. This completes the audit-log write hardening: O_NOFOLLOW (no symlink leaf) + atomic openat traversal (no parent swap) + regular-file (no FIFO/device) + byte-preserving encoding. Shape 389.

omkhar:
@codex review

omkhar:
Rejected hard-linked logs: added an st.Nlink != 1 check right after the S_IFREG check on the Fstat'd fd — a multiply-linked regular file (a hard link to an attacker-chosen inode) is now rejected ('multiply linked'), since a log this helper created and owns has exactly one link. TestAppendAuditLineRejectsHardLink hard-links the log to a victim file and asserts the append errors and the victim is untouched (skips if hard links unavailable). -race green. This completes the audit-log write hardening against a hostile local filesystem: symlink leaf (O_NOFOLLOW) + parent swap (atomic openat) + FIFO/device (O_NONBLOCK+regular) + hard link (Nlink) + byte-faithful encoding.

omkhar:
@codex review

PR #439, opened 2026-07-07
^F C1 (recovery): apple-container session types + recovery/idempotency helpers

omkhar:
**Roadmap C1 — Apple `container` backend evaluation.** Extracts the session types + recovery/idempotency support machinery into their own reviewable PR (below the lifecycle PR that consumes them), keeping each stacked PR under the shape ceiling. All self-contained; no behavior change vs the pre-split lifecycle.

## What (internal/applecontainer/recovery.go + recovery_test.go)
- **Session types:** `StartSessionRequest`, `FinishSessionRequest`, `SessionResult`, `ConformanceTarget`.
- **`lockSession`** — per-session `flock` with a bounded retry over the transient concurrent-`Mkdirat` ENOENT (so the check→heal in Start/FinishSession is atomic).
- **`startTriplet`/`startEventLines`/`startEvent`** — deterministic per-event audit-line renderers (parameterized by timestamp so recovery can re-stamp with the persisted start time).
- **`startedRecordMatchesRequest`** — the recovery record-match gate (full semantically-significant field comparison, timestamps excluded).
- **`auditLineFieldValue`** — exact-token audit-line field parser (same discipline as `auditHasEvents`), used to match `bootstrap_id`/`image_ref` on idempotent retries.

## Validation
`go build`/`vet`/`gofmt`, `go test ./internal/applecontainer/` (recovery unit tests), `go test -race` — all green; 295 lines.

### DIFF
--- a/internal/applecontainer/recovery.go
+++ b/internal/applecontainer/recovery.go
@@ -1,0 +1,218 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+package applecontainer
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"golang.org/x/sys/unix"
+
+	"github.com/omkhar/workcell/internal/host/sessions"
+)
+
+type StartSessionRequest struct {
+	SessionID       string
+	Agent           string
+	Mode            string
+	StartedAt       string
+	Materialization MaterializeResult
+	Bootstrap       BootstrapResult
+}
+
+type FinishSessionRequest struct {
+	Started    SessionResult
+	FinishedAt string
+	ExitStatus string
+}
+
+type SessionResult struct {
+	Record       sessions.SessionRecord
+	RecordPath   string
+	AuditLogPath string
+}
+
+// ConformanceTarget is the session-lifecycle interface the conformance harness
+// drives. It mirrors the four-phase remote-VM lifecycle but with local-VM types.
+type ConformanceTarget interface {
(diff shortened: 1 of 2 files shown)


omkhar:
@codex review

omkhar:
Fixed in a3fb07d (-race green): lockSession now unix.Fchmod's the sid.lock fd to 0o600 right after the O_CREAT open (before flock), so a restrictive umask can't mask the owner bits — same pattern as the record/log Fchmod fixes in merged io. The bounded-retry ENOENT loop is untouched. Test TestRecoveryLockSessionModeUmaskIndependent (non-parallel, umask 0o777, sessions dir pre-created so only the lock's fresh create is under the restrictive umask) asserts mode 0o600 — proven by neutralization (drop Fchmod → mode 0 → FAIL).

omkhar:
@codex review

omkhar:
Fixed in fa5879a (-race green): startedRecordMatchesRequest is now EXHAUSTIVE, not a hand-enumerated field list. It builds the expected record from a single shared startedRecordFields definition (which StartSession also uses for the actual write — so match and write can't drift), seeds started_at/observed_at from the persisted record (excluding only those timestamps), encodes through the writer's own EncodeSessionRecordFrom→DecodeSessionRecord path, and returns expected==current — a full-struct compare. SessionRecord is all comparable scalars, so == covers every current AND future field; a persisted record carrying any unexpected optional metadata (control-plane/log/etc.) the request doesn't set is now rejected. Also DRYed StartSession's inline 21-field write map to the shared definition. Test TestStartSessionRejectsRecordWithExtraField (extra container_name → rejected), proven by neutralization.

omkhar:
@codex review

PR #441, opened 2026-07-07
^B C1 (audit-newline): heal appendAuditLine append boundary (prevents torn-write dead-end)

omkhar:
**Roadmap C1 — Apple `container` backend evaluation.** Fixes a P1 crash-recovery correctness bug in the shared audit-append primitive, surfaced (and empirically reproduced) by an adversarial audit of the lifecycle recovery logic and independently by Codex. Below the lifecycle PR that depends on it.

## The bug
`appendAuditLine` did `io.WriteString(fd, line+"\n")` on an `O_APPEND` fd **without ensuring the file's current tail ends in a newline**. A short/ENOSPC/crash-partial write leaves a newline-less fragment at EOF; the *next* clean append is glued directly onto it, forming one merged physical line that (a) never byte-matches any expected line yet (b) inherits the expected line's trailing field — exactly the heuristic the lifecycle's `conflictingCompleteLine` uses to detect tampering. Result: the recovery path itself **permanently dead-ends the session** with a false tamper verdict, and silently corrupts the append-only evidence log. No attacker required — just the crash/short-write window the recovery code exists to handle.

## The fix
Before writing, `Fstat` the fd for size; if `size>0`, `Pread` the last byte; if it is not `\n`, prepend a separator so the new line lands on its own line instead of merging. The torn fragment is left as a lone inert line (ignored by `filterAuditSessionLines`). The fd is opened `O_RDWR` (was `O_WRONLY`) so the tail read works. Benign shared-log race documented: two concurrent appenders might both prepend → at most an extra empty line, also ignored.

## Validation
`go build`/`vet`/`gofmt`, `go test -race` green. Repro tests `TestAppendAuditLineHealsTornBoundary` + `TestAppendAuditLineHealsMidValueTear` — proven by neutralization (drop the heal → merge → permanent false conflict). remotevm unaffected. 2 files / 77 lines.

### DIFF
--- a/internal/applecontainer/audit.go
+++ b/internal/applecontainer/audit.go
@@ -12,10 +12,20 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"golang.org/x/sys/unix"
 
 	"github.com/omkhar/workcell/internal/host/sessions"
+)
+
+// Bounded non-blocking retry for the shared audit-log append lock: ~50 attempts × 20ms ≈
+// a 1s ceiling. High enough that many genuine concurrent appenders (microsecond critical
+// sections) all win, low enough that an adversarial/stuck advisory-lock holder makes the
+// append fail closed instead of hanging.
+const (
+	auditLockAttempts = 50
+	auditLockBackoff  = 20 * time.Millisecond
 )
 
 // encodeAuditPathValue makes a filesystem path safe to interpolate into a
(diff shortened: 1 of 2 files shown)


omkhar:
@codex review

omkhar:
Fixed in 84f33c0 (-race green): closed the tail-check/append TOCTOU on the shared per-target log. appendAuditLine now takes unix.Flock(leaf, LOCK_EX) right after opening+validating the fd, RE-READS size under the lock (the pre-lock Fstat may be stale), does the tail-Pread + newline-heal + WriteString under the lock, then LOCK_UN after the write (deferred Close is a backstop on error paths). All appenders open the same inode and contend on the same advisory lock → check+append is atomic; no interleaved fragment or duplicate heal. LOCK_EX auto-releases on fd close/process death, so a crashed torn-writer never wedges future appenders (its fragment is healed). Test TestAppendAuditLineConcurrentIntegrity: 48 goroutines append distinct lines to one log under -race after a pre-planted torn fragment → physical-line set == {healed fragment} ∪ {N inputs}, no merged/lost/blank lines. Proven by neutralization: WITH flock 25/25 deterministic pass; WITHOUT 6/10 fail (raced double-heal injects blank lines).

omkhar:
@codex review

omkhar:
Fixed in f97b773 (-race green): the O_RDWR open (added for the tail-Pread) EACCES'd on a pre-existing 0o200 write-only log before the Fchmod could grant read — stranding that log on the crash-recovery path. Now appendAuditLine opens O_WRONLY|O_CREAT first (succeeds for a 0o200 log), validates (regular + Nlink==1) and Fchmod's to 0o600, then re-opens O_RDWR for the flock+tail-Pread+write critical section, re-validating on the new fd. Factored a shared validateAuditLogFd; all opens via openat with O_NOFOLLOW|O_NONBLOCK so symlink/FIFO/hardlink safety holds on every fd. Test TestAppendAuditLineRecoversWriteOnlyLog (0o200 → append succeeds, mode healed to 0o600), proven by neutralization.

omkhar:
@codex review

omkhar:
Fixed in d5e4318 (-race green): the append-serialization flock is now non-blocking with a bounded retry, so a same-UID process holding an advisory LOCK_EX on the shared audit log can no longer hang appendAuditLine forever. LOCK_EX|LOCK_NB in a loop; on EWOULDBLOCK/EAGAIN it backs off (auditLockBackoff=20ms) and retries (auditLockAttempts=50, ~1s ceiling); any other Flock error returns immediately; after the bound it fails-closed ('could not acquire audit log lock: contended') and the caller retries. The real critical section (Fstat+Pread+one WriteString) is microseconds so 48 legitimate concurrent appenders never exhaust the bound. lockSession's blocking flock is untouched (per-session/cooperative, out of scope). Tests: TestAppendAuditLineDoesNotHangOnContendedLock (external LOCK_EX → returns 'contended' in ~1.1s, doesn't hang; neutralize→blocking→hangs past 5s→FAIL) + concurrent-integrity green 8× under -race.

omkhar:
@codex review

PR #443, opened 2026-07-07
^R C1 (session-helpers): extract apple-container session render/compare/validate helpers

omkhar:
**Roadmap C1 — Apple `container` backend evaluation.** Pure refactor: extracts the session-lifecycle helper functions out of `target_session.go` into their own file/PR so the lifecycle orchestration PR stays reviewable and under the shape ceiling. No behavior change.

## What (internal/applecontainer/session_helpers.go + test)
- `conflictingCompleteLine` / `lastFieldKey` — the torn-fragment-tolerant conflicting-audit-line detector (prefix-exclusion) + its field-key helper.
- `startEventsComplete` — exact-complete start-line provenance check (renders expected lines from the persisted record via the shared renderer, token-validates the persisted fields).
- `reconcileStartEvents` — conflict-scan + append-only-missing (shared by fresh-start and recovery).
- `assertRecordContract` — full contract-identity-field drift rejection.
- the `appendAudit` seam var (a moved helper uses it).

## Validation
`go build`/`vet`/`gofmt`, `go test -race` green; new unit tests `TestLastFieldKey`, `TestConflictingCompleteLine`. Behavior-identical to the pre-split lifecycle. remotevm unaffected. 2 files / 189 lines.


omkhar:
@codex review

omkhar:
Implemented the fix in 86dc2c0 (-race green) — on re-assessment my earlier proportionality push-back was wrong about the blast radius, so this is now fixed properly. Verified the per-target audit-LINE format is TARGET-PRIVATE (remotevm has its own independent appendAuditLine/audit lines; no shared consumer parses both; the byte-identical contract is the SessionRecord JSON, untouched) and conformance's validateAuditEvents tolerates extra fields (no exact field-set assertion; TestAppleContainerTargetPassesConformance passes with the sentinel). So the fix is applecontainer-local: added a fixed trailing sentinel field (const auditLineSentinel=' v=1') to every rendered audit line (startEventLines + both finish renders), so the variable identity fields are never last. conflictingCompleteLine's prefix rule now cleanly separates a TORN fragment (strict prefix, cut before the sentinel → ignored/healed, preserving the P1 torn-write DoS avoidance) from a COMPLETE line whose value is merely a prefix of the expected (img vs img:2 → diverges before the sentinel → flagged as CONFLICT). Tests: complete-prefix-value-line-now-flagged + torn-fragment-still-ignored — proven by neutralization (sentinel='' → prefix-value ignored → FAIL); P1 torn-write tests all still pass.

omkhar:
@codex review

PR #444, opened 2026-07-08
^b C1 (fstest-fs): apple-container filesystem-hardening tests

omkhar:
**Roadmap C1 — Apple `container` backend evaluation.** The filesystem-hardening test suite for the (now-merged) apple-container target: symlink/FIFO/hardlink parent-chain guards, atomic create/rewrite/remove (openat+renameat), hardened readers, audit-append rollback, the append-boundary newline-heal + concurrent-integrity + write-only-recovery repro tests, umask-independence. Test-only.

## Validation
`go test -race ./internal/applecontainer/` green; 1 file / 629 lines. All hardening properties proven by neutralization.

### DIFF
--- a/internal/applecontainer/target_session_fs_test.go
+++ b/internal/applecontainer/target_session_fs_test.go
@@ -17,6 +17,62 @@ import (
 	"github.com/omkhar/workcell/internal/host/sessions"
 )
 
+// newSessionFixture builds a target with one materialization ("mid") and one
+// bootstrap ("bid") under a fresh temp state root — the shared preamble of
+// every hostile-fs test here. State root is recoverable via stateRootFor.
+func newSessionFixture(t *testing.T) (AppleContainerTarget, MaterializeResult, BootstrapResult) {
+	t.Helper()
+	ctx := context.Background()
+	target, err := NewAppleContainerTarget(Contract{})
+	mustNil(t, err)
+	state := t.TempDir()
+	mat, err := target.MaterializeWorkspace(ctx, MaterializeRequest{StateRoot: state, TargetID: "tid", MaterializationID: "mid", SourceWorkspace: writeSampleWorkspace(t)})
+	mustNil(t, err)
+	boot, err := target.BootstrapTarget(ctx, BootstrapRequest{StateRoot: state, TargetID: "tid", BootstrapID: "bid", ImageRef: "img:1"})
+	mustNil(t, err)
+	return target, mat, boot
+}
+
+// startReq is the StartSessionRequest literal shared by these tests, parameterized
+// only by the session id.
+func startReq(sid string, mat MaterializeResult, boot BootstrapResult) StartSessionRequest {
+	return StartSessionRequest{SessionID: sid, Agent: "codex", Mode: "strict", StartedAt: "2026", Materialization: mat, Bootstrap: boot}
+}
+
+// mustNotHang fails the test if f blocks past 5s (an unhardened open/read of a
+// FIFO would hang); returns f's error for the caller's own assertions.
+func mustNotHang(t *testing.T, hangMsg string, f func() error) error {
+	t.Helper()
+	done := make(chan error, 1)
+	go func() { done <- f() }()
+	select {
+	case e := <-done:
+		return e
+	case <-time.After(5 * time.Second):
+		t.Fatal(hangMsg)
+		return nil
+	}
+}
+
+// mkfifoOrSkip creates a FIFO at path or skips the test if the runner lacks
+// Mkfifo support.
+func mkfifoOrSkip(t *testing.T, path string) {
+	t.Helper()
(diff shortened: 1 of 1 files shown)


omkhar:
@codex review

omkhar:
Fixed in d1129b6 (-race green): TestWriteSessionRecordAtomicCreateOnce previously passed for the wrong reason — it called writeSessionRecordAtomic(create=true) with an invalid partial map {observed_at:2028}, so EncodeSessionRecordFrom rejected it at required-field validation BEFORE the O_EXCL create-on-existing check ran. Now it uses a VALID SessionRecord map so encode passes and the create path reaches the real O_EXCL step, plus a sanity pre-assertion that the same valid map creates cleanly at a FRESH path — so a failure on the existing path can only come from the create-once guard. Confirmed: neutralizing the create-once guard (allow-existing) makes the test FAIL — previously that regression would've gone undetected.

omkhar:
@codex review

omkhar:
Comprehensive test-validity fix in 05c6066 (-race green). Root cause: several hostile-fs tests put an INVALID record ({version:1}/partial) behind the symlink/hardlink, so on a guard-neutralized path the follow failed at DecodeSessionRecord/encode — the test passed even with the guard removed (false-positive neutralization). Fix: put a VALID encode-passing record behind the object (new writeValidRecordFile helper) and assert the SPECIFIC guard error, so the guard is the ONLY thing that makes the test pass; re-verified by neutralizing the GUARD itself. Audited EVERY hostile-fs test: L426 (RejectsSymlink) + L548 (RewriteReadSymlink) had the decode-masking flaw → valid record + errors.Is(ELOOP); L312 (SymlinkedParent) + 4 others tightened from 'any error' to specific errors (ENOTDIR/ELOOP/'multiply linked'). Each changed test now FAILs when its guard is neutralized (O_NOFOLLOW/Nlink==1/symlink-chain removed). fstest-lifecycle-a/-b have no hostile-fs tests (audited). Full audit table in the report.

omkhar:
@codex review

omkhar:
Fixed in 370390d (-race green): TestWriteSessionRecordAtomicRejectsSymlinkedParent now accepts errors.Is(ELOOP) || errors.Is(ENOTDIR) — an O_NOFOLLOW|O_DIRECTORY open of a symlinked directory component surfaces as ELOOP on Darwin (target runs macOS 26) and ENOTDIR on Linux; both are the O_NOFOLLOW guard rejecting the symlinked parent. Leaf-symlink tests already use ELOOP (unchanged). Guard-neutralization preserved: neutralizing openParentDir's O_NOFOLLOW → the write follows to the real dir and succeeds → error is neither ELOOP nor ENOTDIR → test FAILs (confirmed).

omkhar:
@codex review

omkhar:
Fixed L647 + ran the Simplification Gate in 1ce41c8 (fable-orchestrated, -race green). (L647) TestCreateRecordAtomicUnlinksOnWriteFailure now asserts os.ReadDir(sessions dir) is EMPTY after a forced create-write failure — the dir is created empty and the failing create is its only writer, so a leaked staged temp (sid.json.tmp-<hex>) is now caught; the pre-existing Lstat could NEVER detect it (on the failpoint the rename never runs so sid.json never exists). Proven by neutralization (remove the writeErr-branch Unlinkat → new assert FAILs printing the leaked temp; old Lstat-only would not have). Simplification (behavior-preserving, no guard/coverage loss — verified: every hostile-fs test keeps its valid-record-behind-object shape + specific-error assert + guard-neutralization): shared newSessionFixture (×11) + startReq (×12) + mustNotHang (×5) + mkfifoOrSkip (×5) + validRecordFields helpers; collapsed 3 redundant rewrite tests into one asserting inode-changed+persisted+no-temp; removed dead loop-var shadow. Net −63 lines; 26→24 test funcs (2 deleted were non-security duplicates).

omkhar:
@codex review

omkhar:
Fixed L322 in d6b1ffa (-race green): TestWriteSessionRecordAtomicRejectsSymlinkedParent used create=false, which pre-reads via readFileSafe → openParentDirNoCreate rejected the symlinked parent FIRST, so the WRITE path the name claims was never reached (openParentDir is shared by the read/write call sites; the read reached it first). Switched to create=true (no pre-read) + a FRESH record name behind the symlink, so writeRecordBytesAtomic→openAuditParent's O_NOFOLLOW parent walk is the only possible rejecter — without the guard the create would write THROUGH the symlinked parent and succeed. Kept valid-record discipline + ELOOP||ENOTDIR. Added TestReadSessionRecordSafeRejectsSymlinkedParent to cover the READ call site too. Neutralization: removing openParentDir O_NOFOLLOW → the create=true write-path test FAILs (writes through, error neither ELOOP nor ENOTDIR) — genuinely exercises the writer's parent guard now (masked before under create=false).

omkhar:
@codex review

PR #445, opened 2026-07-08
^b C1 (fstest-lifecycle-a): apple-container input-trust + path-canonicalization tests

omkhar:
**Roadmap C1 — Apple `container` backend evaluation.** The input-trust + path-canonicalization test suite for the (merged) apple-container session lifecycle: contract-field pins, re-derived-not-trusted paths, foreign-target/id-segment rejection, canonicalization (trailing-slash / dot / dotdot / sessions-pin), audit-log rederivation/healing. Test-only. Ran the Simplification Gate (443→386 lines; shared fixtures/helpers reused, zero coverage/assertion change — every guard still proven by neutralization).

## Validation
`go test -race ./internal/applecontainer/` green; 1 file / 386 lines. Every guard proven by neutralization.

### DIFF
--- a/internal/applecontainer/target_session_recovery_test.go
+++ b/internal/applecontainer/target_session_recovery_test.go
@@ -14,21 +14,30 @@ import (
 	"github.com/omkhar/workcell/internal/host/sessions"
 )
 
+// targetRootOf derives the canonical target root from a started session's record
+// path (<targetRoot>/sessions/<id>.json).
+func targetRootOf(started SessionResult) string {
+	return filepath.Dir(filepath.Dir(started.RecordPath))
+}
+
+// rewriteManifest re-persists v at path in writeJSON's on-disk format
+// (MarshalIndent + trailing newline, 0o600) so a self-consistent tamper — the
+// Result and the on-disk bytes edited to agree — still passes the byte-compare.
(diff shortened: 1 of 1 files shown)


omkhar:
@codex review

omkhar:
Fixed L313 in eb16143 (-race green): TestStartSessionRejectsDotDotTargetRoot moved ONLY TargetRoot through ../evil, so ManifestPath still pointed at the original root and StartSession rejected at the manifest-path mismatch — never reaching the ..-canonical/non-canonical-layout check the name claims. Made the fixture SELF-CONSISTENT: mirror the mat/boot manifests (byte-identical) under the sibling dir the .. cleans to (copyManifestToEvil), point mat/boot.TargetRoot at canonicalRoot+'/../evil' and ManifestPath at the mirrored location. Now the manifest checks PASS for the cleaned root, so the canonical-layout assertion (Base('evil') != target id 'tid') is the rejecter; assert the specific 'is not the canonical' error. Neutralization: canonical check disabled → StartSession proceeds to the audit-log check → error is not 'is not the canonical' → FAIL (and proves the fixture reached that far). Demonstrated the OLD fixture PASSES even with the canonical check neutralized (manifest-mismatch fired first) — the finding.

omkhar:
@codex review

PR #455, opened 2026-07-08
.t D7: property-based session-lifecycle tests (testing/quick, guards C1 invariants)

omkhar:
**Roadmap D7 — Test-kind depth.** Adds property-based (randomized-lifecycle) coverage for the C1-hardened session lifecycle. Census: no property lib / benchmarks / bats existed; session lifecycle had only single-transition example tests. This does prong (A) — the highest-value: `internal/host/sessions/lifecycle_property_test.go` generates randomized lifecycles (start → active churn → terminal → post-terminal retries incl. newline-tampered inputs, fixed seed → reproducible) and asserts the invariants the C1 work established:
- **A** no-unusable-record (every encoded byte decodes) · **B** byte-identity/idempotency (re-encode reproduces identical bytes) · **C** terminal-status monotonicity (no regression to non-terminal) · **D** decode→encode fixed point · **E** newline/CR-LF rejection on every required field (log-injection defense) · **F** normalize idempotency.

**Zero new dependency** — used stdlib `testing/quick` (a property lib) rather than adding `rapid`, matching the repo's 3-dep/no-vendor minimal-dep posture. `go.mod`/`go.sum` untouched.

**Proven non-trivial:** each invariant broken in sessions.go → the property CAUGHT it → restored (Inv C: terminal→"stopping" regression accepted; Inv A: trailing-byte; Inv B: per-call nonce; Inv E: dropped newline check). Non-flaky under `-count=5` + `-race -count=3` (fixed seed).

## Validation
gofmt/build/vet clean; `go test ./... -race` green (38 pkgs). Additive test-only; no production code touched.

### DIFF
--- a/internal/host/sessions/lifecycle_property_test.go
+++ b/internal/host/sessions/lifecycle_property_test.go
@@ -1,0 +1,487 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+package sessions
+
+import (
+	"bytes"
+	"math/rand"
+	"reflect"
+	"strings"
+	"testing"
+	"testing/quick"
+)
+
+// Property-based coverage for the session-record lifecycle. The example-based
+// tests in sessions_test.go pin individual transitions; these tests generate
+// randomized lifecycles (start -> active churn -> terminal -> post-terminal
+// retries, including tampered inputs) and assert the invariants the C1
+// hardening established hold for every generated interleaving:
+//
+//   - Idempotency / byte-identity: re-encoding an already-encoded record with no
+//     updates reproduces the exact same bytes (Inv B).
+//   - No unusable record: any bytes EncodeSessionRecordFrom returns decode
+//     cleanly through DecodeSessionRecord (Inv A).
+//   - Terminal-status monotonicity: once a record is terminal, an update that
+//     drives status back to a non-terminal value is refused (Inv C).
+//   - Round-trip fidelity: decode then re-encode is a fixed point (Inv D).
+//   - Newline rejection: injecting CR/LF into any field is always refused, so a
+//     record can never smuggle a forged audit line (Inv E).
+//   - Normalization idempotency: normalize(normalize(x)) == normalize(x) (Inv F).
+//
+// All properties run against a fixed PRNG seed so the explored inputs are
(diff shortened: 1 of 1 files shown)


omkhar:
@codex review

omkhar:
Fixed L272 in a993cb5. The old Inv E reseeded the field pick from len(base) — only 5 distinct lengths (9-13) — so only 5 of 38 newline-checked fields were tampered across 800 iters. Replaced with a deterministic exhaustive TestNewlineRejectionEveryField: all 38 SessionRecord string fields × 4 injections (LF/CR/CRLF/audit-forgery 'x\nsession_id=forged') = 152 sub-tests, each asserting EncodeSessionRecordFrom rejects. No stale hardcoded list: sessionRecordStringFieldTags() enumerates the struct's string fields by reflection + the test asserts base-map keys == exactly those fields BOTH directions (a new string field forces an entry or fails; a removed one is flagged stale). Verified the validator's newline-check block (sessions.go:846-888) lists exactly those 38. Break->fail proven per-field (git_branch/audit_log_path/session_id) + drift guard both directions. go test -race -count=3 green + non-flaky; go.mod unchanged.

omkhar:
@codex review

omkhar:
Fixed both in d436f58 (same 'tolerates errors on must-succeed paths' pattern). Every generated op now carries its expected direction and the property asserts exactly that (no lenient continue): start/active-churn/terminal/terminal-retry = mustSucceed (encode must accept); regression terminal->active = mustFailMonotonic (encode must reject). (L218) on every mustSucceed step, err!=nil is now a property FAILURE naming the step, not a skip. (L241) the round-trip re-encode error is now a failure ('decoded record failed to re-encode'); also expanded recordToUpdates to flatten all 38 string fields (was 24) so the fixed-point check is robust. Regression ops only emitted after a successful terminal op so the monotonicity guard is provably the sole rejecter. Audited the rest of the file — the only other if-err==nil paths are the newline test's must-FAIL assertions (correct, unchanged). Non-trivial both ways: inject a valid-record encoder failure -> 'step 0 (start): valid update was rejected'; make re-encode see an invalid status -> 'decoded record failed to re-encode'. -race -count=3 green + non-flaky; the 152 newline sub-tests still run.

omkhar:
@codex review

PR #457, opened 2026-07-08
^F Add %q audit-line decoder with dup-key rejection (internal/ocsf)

omkhar:
Foundation for **F1 OCSF export** (#456), split out to keep each PR under the size cap.

Adds `decodeAuditLine` — the exact inverse of the repo's authoritative bash `printf %q` audit-record encoder (`bashQuote`, `internal/publishpr/host_exec.go`):
- **`%q` decode:** audit values with spaces/newlines (a space-delimited `ALLOW_ENDPOINTS` allowlist, `argv=hello\ world`) decode to their real value instead of being truncated at the first space by a naive `strings.Fields` split. Un-escapes `\<space>`, ANSI-C `$'…'`, octal.
- **Dup-key rejection (C1 pattern):** a tampered `session_id=A session_id=B` line is refused fail-closed — not silently first/last-wins. Applied to audit *evidence*, so stricter than the first-wins readers is correct.
- **Verified from the authoritative writer:** `TestDecodeAuditLineRoundTripBash` validates against real `bash printf %q` output — not a reimplementation.

Tests: round-trip-vs-bash, deterministic spaced-value decode, ANSI-C newline, unterminated-`$'` fail-closed, torn-token tolerance, dup-key rejection. `go test ./internal/ocsf/` green.

### DIFF
--- a/internal/ocsf/auditline.go
+++ b/internal/ocsf/auditline.go
@@ -166,7 +166,11 @@ func decodeAnsiC(runes []rune, start int) (string, int, error) {
 			default:
 				if isOctal(n) && i+3 < len(runes) && isOctal(runes[i+2]) && isOctal(runes[i+3]) {
 					val := (int(n)-'0')*64 + (int(runes[i+2])-'0')*8 + (int(runes[i+3]) - '0')
-					b.WriteRune(rune(val))
+					// Octal escapes are raw BYTES, not runes: under LC_ALL=C bash %q
+					// emits a non-ASCII value as one \ooo per UTF-8 byte (é → \303\251).
+					// WriteByte reassembles the original multibyte sequence; WriteRune
+					// would re-encode each byte 0xC3/0xA9 as its own rune ("Ã©").
+					b.WriteByte(byte(val))
 					i += 4
 				} else {
 					b.WriteRune(n)
--- a/internal/ocsf/auditline.go
+++ b/internal/ocsf/auditline.go
@@ -1,0 +1,166 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+package ocsf
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Audit records are written by scripts/workcell's append_audit_record_to_path
+// with bash `printf '%q '` per token, so a value containing spaces or control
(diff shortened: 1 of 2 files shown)


omkhar:
@codex review

omkhar:
Fixed L154 in the latest commit. decodeAnsiC now handles ALL bash %q ANSI-C named control escapes — added \a (BEL 0x07), \b (BS 0x08), \e|\E (ESC 0x1b), \f (FF 0x0c), \v (VT 0x0b) alongside the existing \n \r \t \\ \'. Previously these fell through to the default branch and decoded to the literal letter, corrupting any audit value containing those control bytes. Proven by TestDecodeAnsiCNamedControlEscapesRoundTripBash (all 8 named bytes encoded by real bash printf %q round-trip to their real byte) + a bash-independent TestDecodeAnsiCNamedControlEscapes.

omkhar:
@codex review

omkhar:
Fixed L169 in the latest commit. Octal escapes now decode as raw BYTES (WriteByte), not runes. Under LC_ALL=C (session-start locale, scripts/workcell:4279) bash %q emits a non-ASCII value as one octal escape per UTF-8 byte; the old WriteRune(rune(val)) re-encoded each byte as its own rune, so é (\303\251) became mojibake Ã©. Now the raw bytes reassemble the original multibyte UTF-8. Proven by TestDecodeAnsiCOctalBytes (é/café/😀/en-dash) + TestDecodeAuditLineRoundTripBashUTF8LocaleC (a UTF-8 path round-trips through real bash printf %q under LC_ALL=C).

omkhar:
@codex review

omkhar:
Pushing back on the signature finding — it references a commit that does not exist. `git cat-file -t 4069cb66134fd7f25327e7c6d2a7e5c668712d20` → `no such commit`; that SHA is on no branch. All three commits on this PR ARE GPG-signed and GitHub-verified: 8145bc6, ce8a6a3, 2eb0d88 each report `verification.verified=true, reason=valid` via the GitHub API (signed with EDDSA key DA5A8E9F536C42FD, omkhar@gmail.com), and `git log --show-signature` shows 'Good signature' for each. The signed-commits ruleset would block this PR from merging otherwise. This looks like an invented SHA; no change needed.

omkhar:
@codex review

PR #458, opened 2026-07-08
^F A2: deny repo-defined MCP configs by default in strict mode (dated ack to opt in)

omkhar:
**Roadmap A2 — MCP containment.** Census proved (via `--dry-run`) that a committed `.mcp.json` reached the provider **even in strict mode**: `add_shadow_file_mount` materializes the git-index content of tracked control-plane files, and `.mcp.json`/`.github/mcp.json` were in that set — so an untrusted repo could silently wire the agent to arbitrary MCP servers (a boundary/exfil vector). Prior MCP commits (`1f6958d`, `66df9d0`) did not cover repo-defined containment.

## Fix (scripts/workcell)
- **Deny-by-default:** in every non-breakglass mode, `.mcp.json` and `.github/mcp.json` get a neutralized `{}` (no servers) overlaid via `add_denied_mcp_file_mount`, so repo-defined servers never reach the provider.
- **Fail-closed:** the repo content (committed or working-tree, parseable or not) is replaced wholesale — a malformed config cannot fall open.
- **Layered/independent:** the deny holds even under `--allow-control-plane-vcs` (acknowledging Git visibility does not lift it).
- **Ack path:** `--allow-repo-mcp` + dated `--ack-repo-mcp=YYYY-MM-DD` (mirrors the `validate_ack_token` pattern used by `--ack-arbitrary-command`, stronger than a boolean — appropriate for an exfil surface; rejects missing/stale tokens + redundant-in-breakglass).
- **Recorded:** `workspace_repo_mcp=denied|acknowledged` in the typed SessionRecord, host audit log, container env, diagnostics.

## Tests (verify-invariants.sh, +76 lines)
Deny-by-default; **load-bearing proof** = the repo MCP surface is NOT a mount source without the ack; layered deny under control-plane-vcs; acknowledged path mounts the real config; missing-ack / stale-token / breakglass-redundant rejections; fail-closed on malformed config. The block was extracted verbatim + run against the real launcher — all assertions pass.

## Contract lockstep
man page, docs/stability-contract.md, docs/adapter-control-planes.md, help text, policy/public-contract.toml, typed SessionRecord field (+ tests), control-plane-manifest regenerated. All verifies pass (manifest/operator-contract/public-contract green locally).

Follow-up (noted, out of scope): Codex `.codex/mcp/config.toml` arrives via the whole-`.codex` dir mount; a directory-scoped carve-out is deferred (documented). CI runs the full verify-invariants.sh (the worktree sandbox tripped the launcher's sensitive-path guard; the block was validated in isolation).

### DIFF
--- a/docs/adapter-control-planes.md
+++ b/docs/adapter-control-planes.md
@@ -145,3 +145,21 @@ acknowledged, audited, and tested.
 Workcell ships no live MCP defaults in the adapter baselines. Reviewed MCP
 state must arrive through explicit operator inputs, not ambient workspace
 content.
+
+Repo-defined MCP server-definition surfaces (`.mcp.json`, `.github/mcp.json`)
+are classified `deny` in every non-breakglass mode: an untrusted workspace can
+otherwise wire the agent to arbitrary MCP servers, a boundary/exfil risk. The
+launcher overlays a neutralized, schema-valid empty config (`{"mcpServers": {}}`,
+matching the shipped `adapters/claude/mcp-template.json` baseline) over each
+surface so the repo-defined servers never reach the provider process while the
+config stays parseable, and it fails closed — an unparsable or malformed
+workspace config is replaced wholesale rather than honored. This deny is independent of `--allow-control-plane-vcs`: acknowledging
+control-plane Git visibility does not lift it.
+
+An operator opts a workspace's MCP configs back in with the explicit,
+`ack-required` pair `--allow-repo-mcp` plus a dated `--ack-repo-mcp=YYYY-MM-DD`.
+The launcher records the decision as `workspace_repo_mcp=denied` or
+`workspace_repo_mcp=acknowledged` in the session record and the host audit log.
+Codex's `.codex/mcp/config.toml` arrives through the whole-`.codex` directory
+surface and remains governed by the existing control-plane masking; a
+directory-scoped carve-out for it is tracked as follow-up work.
--- a/docs/invariants.md
+++ b/docs/invariants.md
@@ -99,6 +99,7 @@ Examples:
 - `development`
 - package mutation inside a mutable container
 - `--allow-control-plane-vcs`
+- `--allow-repo-mcp`
 - `--allow-arbitrary-command`
 - `breakglass`
 - host-side debug or transcript capture
--- a/docs/stability-contract.md
+++ b/docs/stability-contract.md
@@ -80,8 +80,9 @@ Experimental or explicitly gated (may change; already marked in `--help`):
 - `--ui gui` (not implemented; fails closed).
 - preview targets `aws-ec2-ssm` and `gcp-vm`.
 - `--mode breakglass` (gated behind a dated `--ack-breakglass`),
-  `--allow-arbitrary-command` (behind `--ack-arbitrary-command`), and
-  `--allow-control-plane-vcs` (behind `--ack-control-plane-vcs`).
+  `--allow-arbitrary-command` (behind `--ack-arbitrary-command`),
+  `--allow-control-plane-vcs` (behind `--ack-control-plane-vcs`), and
+  `--allow-repo-mcp` (behind a dated `--ack-repo-mcp`).
 
 ## Stable machine-readable output
 
@@ -143,7 +144,8 @@ mechanism.
 `session_audit_dir`, `audit_log_path`, `debug_log_path`,
 `file_trace_log_path`, `transcript_log_path`, `started_at`, `observed_at`,
 `finished_at`, `exit_status`, `initial_assurance`, `current_assurance`,
-`final_assurance`, `workspace_control_plane`, `bootstrap_id`, `image_ref`.
+`final_assurance`, `workspace_control_plane`, `workspace_repo_mcp`,
+`bootstrap_id`, `image_ref`.
 
 `SessionExport` (`session export`) wraps this in `session` and adds
 `audit_records`.
--- a/internal/host/sessions/lifecycle_property_test.go
+++ b/internal/host/sessions/lifecycle_property_test.go
@@ -311,6 +311,7 @@ var newlineBaseUpdates = map[string]string{
 	"current_assurance":       "managed-mutable",
 	"final_assurance":         "managed-mutable",
 	"workspace_control_plane": "masked",
+	"workspace_repo_mcp":      "denied",
 	"bootstrap_id":            "boot-1",
 	"image_ref":               "img@sha256:aaa",
 }
@@ -481,6 +482,7 @@ func recordToUpdates(r SessionRecord) map[string]string {
 	add("initial_assurance", r.InitialAssurance)
 	add("final_assurance", r.FinalAssurance)
 	add("workspace_control_plane", r.WorkspaceControlPlane)
+	add("workspace_repo_mcp", r.WorkspaceRepoMcp)
 	add("bootstrap_id", r.BootstrapID)
 	add("image_ref", r.ImageRef)
 	return updates
--- a/internal/host/sessions/sessions.go
+++ b/internal/host/sessions/sessions.go
@@ -54,6 +54,7 @@ type SessionRecord struct {
 	CurrentAssurance      string `json:"current_assurance,omitempty"`
 	FinalAssurance        string `json:"final_assurance,omitempty"`
 	WorkspaceControlPlane string `json:"workspace_control_plane,omitempty"`
+	WorkspaceRepoMcp      string `json:"workspace_repo_mcp,omitempty"`
 	BootstrapID           string `json:"bootstrap_id,omitempty"`
 	ImageRef              string `json:"image_ref,omitempty"`
 }
@@ -99,6 +100,7 @@ func SessionDiffMetadataLines(record SessionRecord) []string {
 		fmt.Sprintf("container_name=%s", record.ContainerName),
 		fmt.Sprintf("observed_at=%s", record.ObservedAt),
 		fmt.Sprintf("workspace_control_plane=%s", record.WorkspaceControlPlane),
+		fmt.Sprintf("workspace_repo_mcp=%s", record.WorkspaceRepoMcp),
 	}
 }
 
@@ -248,6 +250,7 @@ func SessionShowLines(record SessionRecord) []string {
 		fmt.Sprintf("file_trace_log_path=%s", record.FileTraceLogPath),
 		fmt.Sprintf("transcript_log_path=%s", record.TranscriptLogPath),
 		fmt.Sprintf("workspace_control_plane=%s", record.WorkspaceControlPlane),
+		fmt.Sprintf("workspace_repo_mcp=%s", record.WorkspaceRepoMcp),
 	}
 }
 
@@ -399,6 +402,8 @@ func encodeSessionRecordFrom(existing []byte, updates map[string]string, source 
 			record.FinalAssurance = value
 		case "workspace_control_plane":
 			record.WorkspaceControlPlane = value
+		case "workspace_repo_mcp":
+			record.WorkspaceRepoMcp = value
 		case "bootstrap_id":
 			record.BootstrapID = value
 		case "image_ref":
@@ -883,6 +888,7 @@ func validateSessionRecord(record SessionRecord, source string) error {
 		{name: "current_assurance", value: record.CurrentAssurance},
 		{name: "final_assurance", value: record.FinalAssurance},
 		{name: "workspace_control_plane", value: record.WorkspaceControlPlane},
+		{name: "workspace_repo_mcp", value: record.WorkspaceRepoMcp},
 		{name: "bootstrap_id", value: record.BootstrapID},
 		{name: "image_ref", value: record.ImageRef},
 	} {
--- a/internal/host/sessions/sessions_test.go
+++ b/internal/host/sessions/sessions_test.go
@@ -123,6 +123,7 @@ func TestSessionDiffMetadataLines(t *testing.T) {
 		ContainerName:         "workcell-session-1",
 		ObservedAt:            "2026-04-08T12:00:30Z",
 		WorkspaceControlPlane: "masked",
+		WorkspaceRepoMcp:      "denied",
 	})
 
 	want := []string{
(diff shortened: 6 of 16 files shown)


omkhar:
@codex review

omkhar:
Fixed L7552 in 2c97fd3. My acked path had two mount branches; only the control-plane-vcs branch was correct. The masked branch (acked, no --allow-control-plane-vcs) routed the MCP surface through add_shadow_file_mount (git-index-only; placeholder-masks untracked), so an acknowledged working-tree .mcp.json was silently placeholder-masked (untracked) or served stale (index vs working-tree) while the record said acknowledged. Fix: in that branch, an ACKNOWLEDGED MCP surface now routes to register_workspace_control_plane_file_mount (same helper the vcs branch uses) → mounts the live working-tree file read-only, refusing symlinks. Makes workspace_repo_mcp=acknowledged TRUE without coupling to --allow-control-plane-vcs (chose this over require-the-flag for better UX). Deny-by-default (no-ack) path UNCHANGED. 2 new load-bearing tests: acked+untracked → real file is the mount source (not placeholder); acked+tracked-locally-modified → live workspace file, not a shadow index snapshot — both proven load-bearing (revert Branch 2 → assertions fail). go test -race green; shellcheck/manifest/operator-contract/public-contract pass.

omkhar:
@codex review

omkhar:
Fixed the P1 symlink bypass in 7133f41. A workspace .mcp.json symlinked to an absolute in-container path (e.g. /workspace/evil.json) is dangling on the HOST (-e false) but resolves IN-CONTAINER, so gating the deny presence test on -e alone left the symlink exposed via the bind mount. Audited ALL branches: (1) the --allow-control-plane-vcs deny branch (the named one) gated on -e || git-index and MISSED -L → fixed to -e || -L || git-index; (2) the masked deny branch was already -f || -L || git-index (correct); (3) register_workspace_control_plane_file_mount (shared by the acked branches + all control-plane files under vcs) ran [[ -e ]] || return 0 BEFORE its -L refuse, so a dangling symlink silently returned no-mount → moved the -L symlink refusal BEFORE the -e gate so it fails closed. Symlink targets are never resolved/mounted. New load-bearing tests: dangling-on-host symlinked .mcp.json + .github/mcp.json through both the masked and control-plane-vcs paths assert the deny overlay masks the surface, the evil target is never a mount source, and state=denied; plus an acked-symlink-refused test. Proven load-bearing (reverted -L → reproduced the bypass). All prior A2 tests green; shellcheck/manifest/operator-contract/public-contract pass.

omkhar:
@codex review

omkhar:
Fixed both in 6935363. FIX1 (P1 parent-symlink traversal): register_workspace_control_plane_file_mount only rejected a symlinked LEAF; a symlinked PARENT (.github -> /outside) leaves .github/mcp.json a regular file reached THROUGH the parent symlink → Docker follows the host source out of the workspace. Reused the launcher's existing per-component -L scan idiom (reject_symlinked_host_path_components style; no realpath, no macOS portability risk) as refuse_symlinked_workspace_parent_component, applied to ALL branches: register_ (acked + control-plane-vcs files) confines the whole relative path (refuse symlinked parent, then leaf, before the existence gate); both deny branches call it before add_denied_mcp_file_mount (symlinked parent can't be reliably masked → fail closed). Load-bearing proven IN-PLACE (reverting reproduces the .github/mcp.json→/outside leak; note: my earlier /tmp-copy revert proofs were invalid since a copied script can't source assurance.sh via relative ROOT_DIR, so I redid this in-place). FIX2 (P2 assurance): session_assurance_initial now returns lower-assurance-repo-mcp when repo-mcp is acknowledged (same mechanism/precedence as control-plane-vcs/persistent-cache) + a matching notice; deny-by-default keeps managed-mutable. New load-bearing tests + lockstep docs (man/invariants/requirements/assurance.sh notice) + manifest regen. go test -race green; shellcheck/manifest/op-contract/public-contract pass.

omkhar:
@codex review

omkhar:
Fixed L2697 in 2edc3fb. With both --allow-control-plane-vcs and --allow-repo-mcp, session_assurance_initial's vcs branch returned early → the repo-mcp reason was dropped from the label and its notice never fired. No existing multi-reason convention existed (final assurance overrides via package-mutation; the notice case emitted one arm), so I introduced a deterministic +-joined label: session_assurance_initial builds an ordered reasons array and emits lower-assurance-<r1>+<r2> (vcs then repo-mcp). Single/no-reason paths are byte-for-byte unchanged (vcs-only, mcp-only, managed-mutable, persistent-cache). emit_session_assurance_notice now matches each reason as a substring and fires every applicable warning (emitted-flag set once), so the combined label emits BOTH notices while each single reason still emits exactly one. Load-bearing: both flags → lower-assurance-control-plane-vcs+repo-mcp (stderr+env) + both warnings; notice unit test asserts both fire combined / one each single / zero for full. man page documents the combined label; manifest regen; the + round-trips through the typed SessionRecord. go test -race green; shellcheck/manifest/op-contract/public-contract pass.

omkhar:
@codex review

omkhar:
Fixed L7378 regression in 57d0eae. My parent+leaf symlink refusal ran BEFORE the existence check in register_workspace_control_plane_file_mount (called unconditionally for optional surfaces like .github/copilot-instructions.md), so a workspace legitimately keeping .github as a symlink was refused solely because an absent optional file's parent is a symlink (same bug in the Branch-1 deny block). Reorder: presence determined FIRST — present = [ -e $src ] || [ -L $src ] || git_index_has_path; not present → return 0 (absent optional surface, no block); present → confine (refuse symlinked parent, then leaf) and mount. The -e||-L||git-index test keeps the earlier fixes intact: L7523 (dangling leaf symlink, -e false -L true → present → neutralized), L7560 (file present through symlinked parent, -e true → parent refusal fires), regression (truly absent optional → early return). Load-bearing: symlinked .github + absent optional files + --allow-control-plane-vcs → launches, nothing mounted (proven: stashing the fix → blocks with 'parent component is a symlink'). L7523/L7560/L2697 all still green. go test -race green; shellcheck/manifest/op-contract/public-contract pass.

omkhar:
@codex review

omkhar:
Fixed L7052 in 1d74a56. add_denied_mcp_file_mount overlaid bare {} (schema-empty) — a provider expecting the mcpServers map could treat it as broken, turning 'repo merely contains .mcp.json' into a parse problem. Added a centralized helper denied_mcp_config_json <rel> returning the canonical zero-server config, piped instead of {}. .mcp.json → {"mcpServers": {}} (byte-equivalent to the repo's shipped adapters/claude/mcp-template.json baseline, which I read: it's exactly that). .github/mcp.json → same map (Copilot's repo MCP config uses the same top-level mcpServers; the *) default also falls closed to it). Zero servers reach the agent AND the file is parseable. Tests: overlay content == {"mcpServers": {}} for both surfaces, jq-valid empty map (zero servers), byte-equiv to the baseline; existing 'repo servers never reach provider' assertions green. Doc lockstep (adapter-control-planes.md). All prior A2 guarantees green; shellcheck/manifest/op-contract/public-contract pass.

omkhar:
@codex review

omkhar:
CI fix: codespell flagged 'unparseable' → 'unparsable' in docs/adapter-control-planes.md + scripts/verify-invariants.sh (both the Spelling-and-manpage and Validate-repository jobs run codespell). Fixed both; codespell now clean across all changed files, manifest/operator-contract/public-contract verify pass, go build green. No behavior change.

omkhar:
@codex review

omkhar:
CI fix #2: the Validate-repository job enforces shellcheck at info severity, and my assurance-notice unit-test harness had an SC2028 on verify-invariants.sh:5985 (echo with a \\n escape). Refactored the emit from `echo "...\\n..."` to `printf '%s\\n' "..."` (SC2028 is echo-specific; printf doesn't interpret escapes in its %s argument, so the generated stub 'workcell_runtime_state_value() { printf %s\\n ...; }' is byte-identical). shellcheck now clean at all severities on scripts/workcell, verify-invariants.sh, assurance.sh; the combined-notice test still passes. No behavior change.

omkhar:
@codex review

omkhar:
CI fix #3: the Validate job also runs shfmt (-ln=bash -i 2 -ci); my grep_repo_mcp test helper was a one-liner that shfmt reformats to multi-line. Applied shfmt with the repo's exact flags (only that one hunk changed — verified no whole-file reformat). shfmt -d + shellcheck now both clean on verify-invariants.sh. No behavior change.

omkhar:
@codex review

omkhar:
Fixed L7631 in 8373399. In default isolated-workspace mode the session runs in a session-owned CLONE (WORKSPACE), but the acked-MCP branch mounted ${workspace}/${rel} from WORKSPACE_SHADOW_SOURCE (the original checkout) — so a post-launch edit to the source would change the running isolated session's MCP config, breaking isolation. Added a session_workspace param (=${WORKSPACE}) to prepare_workspace_control_plane_shadow; the ACKED MCP mount now uses register_workspace_control_plane_file_mount "${session_workspace}" (the clone, same source as the whole-workspace /workspace bind mount) in both branches, while deny/masking keep the source (the deny overlay is a source-independent {"mcpServers": {}}). Isolated: committed config comes from the clone (insulated); an untracked source file isn't in the clone (consistent with commit-only isolation). Direct mode: session_workspace==source, unchanged. Load-bearing: isolated dry-run → /workspace binds the clone, source .mcp.json/.github/mcp.json NOT mount sources (stash-to-pre-fix leaks them). Deny still neutralizes; symlink parent/leaf confinement runs on the clone path. Ran the exact CI validators: shfmt -ln=bash -i 2 -ci = 0 hunks, shellcheck -x = 0, codespell = 0; go test -race green.

omkhar:
@codex review

omkhar:
Fixed both in 4609720. (L2698) --allow-repo-mcp now marks EXECUTION_PATH=lower-assurance-repo-mcp (mirrors control-plane-vcs), combined +-joined when both acks set (lower-assurance-control-plane-vcs+repo-mcp); deny-by-default keeps managed-tier1. EXECUTION_PATH feeds one var consumed by the dry-run line, write_session_record_initial, and append_launch_audit_record, so all three surfaces are consistent (load-bearing: stash-to-pre-fix showed repo-mcp-only was managed-tier1). man page documents the values. (L2856) workspace_repo_mcp is a genuine audit-line key (append_launch_audit_record:2856 + append_exit_audit_record:2897), was absent from knownAuditFields so OCSF bucketed the deny/ack decision as redacted-unknown — added it to internal/ocsf/auditline.go (alphabetical) + the auditfields_test.go positive list; the documented regeneration recipe now reproduces the map exactly (empty comm diff both ways). go test ./internal/ocsf green. Exact CI validators all zero (shfmt -ln=bash -i 2 -ci, shellcheck -x, codespell); go test -race green; contracts pass.

omkhar:
@codex review

omkhar:
Fixed L7647 in 238a7ec. My L7631 fix sourced the acked MCP mount from the isolated clone (session_workspace); a real launch creates the clone first so it mounts, but --dry-run sets WORKSPACE to the future clone path WITHOUT creating it, so the probe hit a nonexistent path → no mount → dry-run falsely implied the acked config won't reach the provider. New helper register_acked_repo_mcp_mount(source, session_workspace, rel), matching the existing masked-path pattern (add_shadow_file_mount probes the git index): if session_workspace exists (-d) — real launch / non-isolated — probe+mount it directly (unchanged); if not (isolated dry-run phantom) — decide existence via git_index_has_path(source) (a clone contains committed regular files; untracked source files aren't cloned), apply symlink parent/leaf confinement against the source, and register with the CLONE path displayed. The -d check cleanly separates real-clone from dry-run-phantom. Load-bearing: isolated dry-run + committed .mcp.json + .github/mcp.json → both shown mounted from the clone path (stash-to-pre-fix showed no mount). Untracked-in-isolated is unreachable (isolated refuses a dirty workspace). Non-isolated + deny unchanged. Exact CI validators all zero (shfmt -ln=bash -i 2 -ci, shellcheck -x, codespell); go test -race green.

omkhar:
@codex review

omkhar:
Rebased onto main after F1 PR #461 (OCSF session-events) merged — the two share the session help/man/manifest surfaces. Conflicts resolved: (1) session help combines '--format ocsf maps the record and audit lifecycle events' with the accurate 'credentials, tokens, and operator home paths' redaction scope + the free-form-withheld note; (2) control-plane-manifest regenerated. Plus: F1's drift guard (TestOCSFCoversEverySessionRecordField) requires every SessionRecord field appear in the OCSF export, so my new workspace_repo_mcp SessionRecord field needed session.workspace_repo_mcp added to sessionEvent — done (golden regenerated; field is omitempty so an empty value stays out). All green: ocsf test, manifest/operator-contract/public-contract verify, shfmt -ln=bash -i 2 -ci = 0, shellcheck -x = 0, codespell = 0, go test -race.

omkhar:
@codex review

omkhar:
Fixed L7460 in e721536. My L7647 isolated-dry-run branch resolved existence via git_index_has_path but kept an on-disk -f short-circuit; under sparse-checkout/skip-worktree a config is TRACKED in the index but absent on disk, so -f skipped the future-clone mount even though a real isolated launch clones + checks the tracked file out. Replaced the -f/-L on-disk checks with a git-index-MODE decision: git_index_path_mode(source, rel) via ls-files --stage → not tracked = no mount; mode 120000 (symlink) = REFUSE (the symlink-index case); 100644/100755 (regular blob) = register the clone-path mount; other modes = not mountable. Real-launch path (clone exists, -d) still probes the actual clone; non-isolated + deny unchanged. Load-bearing: sparse/skip-worktree fixture (update-index --skip-worktree + rm, clean workspace) → mount shown from the clone path (stash-to-pre-fix skipped it); tracked-symlink index entry → refused. Exact CI validators all zero (shfmt -ln=bash -i 2 -ci, shellcheck -x, codespell); go test -race green; manifest/operator-contract/public-contract pass.

omkhar:
@codex review

omkhar:
@codex review

PR #459, opened 2026-07-08
^F Add per-writer encoding selection to the audit-line decoder (internal/ocsf)

omkhar:
Foundation for the **F1 OCSF export** (#456), split out to keep each PR under the size cap (it consumes this selection).

The audit log has **two** on-disk encoders:
- `scripts/workcell` writes tokens with bash `printf %q` (all launcher backends: colima/docker-desktop/aws-ec2-ssm/gcp-vm).
- `internal/applecontainer` writes plain `Sprintf` lines and percent-encodes **only** its path fields via `encodeAuditPathValue` — a literal backslash in a POSIX path (`/tmp/a\b`) is emitted **unchanged**.

Decoding an apple-container record with the `%q` decoder dropped backslashes (`/tmp/a\b` → `/tmp/ab`), corrupting the path evidence (this is the F1 L369 finding).

- `decodeAuditLine` now takes an `auditEncoding`; `auditEncodingForProvider(provider)` selects **percent** decoding **only** on the unambiguous `apple-container` provider (never a launcher backend, so the two encoders never share a log), bash `%q` as the safe default.
- Only the AppleContainer-encoded path fields are percent-decoded, so a raw `%` in a `%q` record isn't corrupted.
- Exports the canonical inverse as `applecontainer.DecodeAuditPathValue` (thin wrapper over the existing unexported decoder — reused, not reimplemented).
- Dup-key rejection + full ANSI-C/octal `%q` handling unchanged.

Tests: backslash-preserved apple-container decode with a **control** proving `%q` corrupts it; `%q` spaces/named/octal unchanged; dup-key rejection under both encodings. `go test ./internal/ocsf/ ./internal/applecontainer/` green.

### DIFF
--- a/internal/applecontainer/audit.go
+++ b/internal/applecontainer/audit.go
@@ -50,6 +50,15 @@ func encodeAuditPathValue(value string) string {
 	return b.String()
 }
 
+// DecodeAuditPathValue is the exported canonical decoder for audit path values
+// this package writes with encodeAuditPathValue. Cross-package readers (the OCSF
+// export in internal/ocsf) reuse it to recover the exact original path from a
+// percent-encoded AppleContainer audit field, so a literal backslash in a legal
+// POSIX path survives instead of being mis-read as a shell escape.
+func DecodeAuditPathValue(value string) string {
+	return decodeAuditPathValue(value)
+}
+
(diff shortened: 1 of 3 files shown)


omkhar:
@codex review

PR #460, opened 2026-07-08
^F Add known-audit-field allowlist for safe OCSF unmapped keys

omkhar:
Foundation for **F1 OCSF export** (#456), split out to keep each PR under the size cap.

Audit records are parsed from mutable text and arbitrary non-duplicate keys are accepted, so a tampered line like `session_id=s ghp_TOKEN=x` could make a secret-shaped field **name** into a JSON property. `knownAuditFields` is the set of field names the writers actually emit (`scripts/workcell` `append_*` records + `internal/applecontainer` Go writer + framing keys), derived from the writers (regeneration documented on the var).

The OCSF mapping (F1) turns **only** these into typed `audit.<key>` properties; any other key is bucketed under one fixed **redacted** property, so a secret can never become a property name and a new legitimate writer key degrades safely to the bucket rather than leaking.

Test pins the allowlist against writer keys and asserts untrusted keys (`ghp_…`, a home path, empty) are NOT allowlisted. `go test ./internal/ocsf/` + `go vet` green.

### DIFF
--- a/internal/ocsf/auditfields_test.go
+++ b/internal/ocsf/auditfields_test.go
@@ -1,0 +1,35 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+package ocsf
+
+import "testing"
+
+// TestKnownAuditFieldsCoversWriterKeys pins the allowlist against the audit-field
+// names the writers actually emit. The OCSF mapping turns ONLY these into typed
+// audit.<key> properties; any other key is bucketed under a single redacted
+// property, so a tampered audit line cannot make a secret-shaped key into a JSON
+// property name. If a writer gains a new field, add it here (a missing entry
+// degrades safely to the redacted bucket rather than leaking).
+func TestKnownAuditFieldsCoversWriterKeys(t *testing.T) {
+	// A representative subset the launcher + apple-container writers emit on
+	// every record; the mapping depends on these being recognized.
+	expected := []string{
+		"event", "session_id", "timestamp", "ts",
+		"target_kind", "target_provider", "target_id",
+		"exit_status", "workspace", "workspace_origin",
+		"record_digest", "prev_digest",
+	}
+	for _, k := range expected {
+		if _, ok := knownAuditFields[k]; !ok {
+			t.Errorf("knownAuditFields is missing writer-emitted key %q", k)
+		}
+	}
+	// A tampered/secret-shaped key must NOT be in the allowlist (so it routes to
+	// the redacted bucket instead of becoming a typed property name).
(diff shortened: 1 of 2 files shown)


omkhar:
@codex review

omkhar:
Fixed all 3 in the latest commit. (L82/L173) Added `v` — auditLineSentinel (" v=1", session_helpers.go:23) is appended to every apple-container audit line, so it's a legit schema-version field; without it apple-container records sent their version token to the redacted bucket. (L126) Removed all 7 git_* entries: git_dir/work_tree/common_dir/exec_path are core.*/git_* alias-guard substrings (scripts/workcell:6775-6778), and git_branch/git_head/git_base are SessionRecord fields written by write_session_record_initial (line ~2946) — NONE are emitted by the audit-append writers (verified: 0 emissions in any audit line). Allowlisting them let a tampered git_dir=secret line become a typed unredacted audit.git_dir; now they route to the redacted bucket. Fixed the regeneration recipe to scope to the append_*_audit_record functions (not all of scripts/workcell) + add the sentinel v — the corrected recipe emits v and no git_* (verified). Test now asserts v is allowlisted and every git_* is NOT.

omkhar:
@codex review

omkhar:
Fixed L99 (+ the whole allowlist accuracy) in 3c7c71b. Re-derived from ALL audit emit paths: the 4 bash audit-writer bodies (launch/exit/assurance_change/session_control), the per-event inline keys at every append_session_control_audit_record CALL SITE (source/command/argv/force/transport_status/final_assurance/stdin_mode/container_name/...), the append_audit_record_to_path framing (timestamp/prev_digest/record_digest), the apple-container key=%s writers, and the v sentinel. 69 keys (was 85). 16 REMOVED — each verified (grep) to be written ONLY by write_session_record (go_hostutil session-record-write, the session RECORD) or echoed to stderr, never an audit-log writer: audit_log_path, debug_log_path, file_trace_log_path, transcript_log_path, session_audit_dir, target_assurance_class, workspace_root, worktree_path, monitor_pid, current_assurance, initial_assurance, live_status, observed_at, started_at, finished_at, prepare. (On apple-container lines started_at/finished_at are the ts= VALUE, not keys.) The recipe was rewritten to cover all four bodies + call sites (the old awk missed assurance_change, so it couldn't even reproduce 'reason') and NOT sweep write_session_record/git_* guards — the literal comment recipe now reproduces the 69-key map exactly (empty diff, confirmed). Test asserts positives across all 5 emit paths + negatives for all 16 removed names + git_*/secret-shaped. Follow-up: once A2 (#458) lands its workspace_repo_mcp audit key, add it (degrades safely to the redacted bucket meanwhile). go build/vet/test green.

omkhar:
@codex review

PR #461, opened 2026-07-08
^F F1 (1/2): OCSF session-event export (workcell session export --format ocsf)

omkhar:
**Roadmap F1 — OCSF export, part 1 of 2** (split by event type to stay under the size cap; the decoder/per-writer/allowlist primitives already merged as #457/#459/#460).

Adds the OCSF export skeleton: `workcell session export --format json|ocsf` maps the **session record** to one OCSF **Application Lifecycle** event (class_uid 6002, category 6) as JSON Lines. Types (Product/Metadata/App/Device/Event), Export/WriteJSONL/mapExport (session-only here — the per-audit-record events come in part 2, #f1a), versioning (`metadata.version` 1.3.0 + `mapping_version`), and the shared G2 redactor on all free-form strings.

`Export` keeps its `([]Event, error)` signature (error reserved for part 2) so the CLI is stable across the stack. Full contract lockstep: CLI flag, session help, man page, operator-contract, control-plane manifest, and the scenario's OCSF assertions (class_uid 6002, category_uid 6, app.uid, JSONL-not-bundle).

## Validation
`go test ./... -race` green (session classification, versioning, JSONL well-formedness, determinism, session redaction leak-proof, SessionRecord field-parity drift guard, golden); operator-contract + manifest verify pass; 825 lines.

Part 2 (**f1a/ocsf-audit**, #TBD) stacks on this to add per-audit-record events + the audit-field allowlist bucketing + free-form hard-redaction.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #462, opened 2026-07-08
^F F1 (2/2): OCSF per-audit-record events + free-form redaction

omkhar:
**Roadmap F1 — OCSF export, part 2 of 2** (stacks on #461 session-events, now merged). Adds per-audit-record OCSF events to `workcell session export --format ocsf`.

- Iterates the session's audit records → one OCSF Application Lifecycle event per line (started/launch→Start, finished/exit→Stop, else→Other), on top of #461's session-summary event.
- **Per-writer audit decode** (bash `%q` vs AppleContainer percent-encoding, from the merged #459), **dup-key rejection** fail-closed, **device/app from the authoritative SessionRecord** (audit lines are sparse), **monitor_pid parity**.
- **Field-name allowlist** (merged #460 `knownAuditFields`): unexpected/tampered keys bucket under one **redacted** property, never a typed key.
- **Free-form `argv` hard-redaction (P1)**: the operator's detached-session message is replaced with a fixed placeholder — regex redaction can't sanitize arbitrary prose (a `--password hunter2` in a message would otherwise survive). Help/man updated: free-form payloads are withheld.

## Validation
`go test ./... -race` green; golden regenerated to the 3-event shape (each event's `app` carries the OCSF-required `vendor_name` from #461). Rebased on merged #461 (help text combined: "record and audit lifecycle events" + the accurate "credentials, tokens, and operator home paths" scope + the free-form-withheld note).


omkhar:
@codex review

omkhar:
Fixed all 5 in 17f48bc. (L441 P1) Unexpected-key VALUES are now hard-redacted to the FreeFormPlaceholder (key stays regex-redacted) — the regex redactor can't sanitize split-CLI/prose secrets, same as argv. (L467 P1) Added knownAuditEvents (the bounded enum the writers emit); the OCSF message echoes event only when recognized, else the safe generic 'workcell audit Other' and the raw event is routed to the hard-redacted bucket — a tampered event=deploy --password hunter2 no longer leaks. (L410) A torn event=session_finished with no exit_status now maps to Unknown, not Success (statusFromExit("") no longer fabricates a clean outcome); exit_status=0 stays Success, non-zero → Failure. (L198) Bumped MappingVersion 1→2 (single-summary-event → multi-event stream), golden + doc updated. Load-bearing tests: unexpected-value + tampered-event → no hunter2/--password in output (precondition proves regex alone doesn't mask); torn→Unknown, complete-exit0→Success; every event carries mv=2. Exact CI validators zero (shfmt -ln=bash -i 2 -ci, shellcheck -x, codespell); go test -race green.

omkhar:
@codex review

omkhar:
Fixed both in e0ce1ef. (L448 P1) The prior fix hard-redacted the unexpected VALUE but left the KEY regex-redacted; a tampered deploy --password hunter2=x decodes that prose as the KEY, which regex can't sanitize → leak in a property name. Now BOTH key and value are hard-redacted: audit.unexpected_fields = '[REDACTED-FREEFORM]=[REDACTED-FREEFORM]' (zero attacker-controlled text; the entry count still signals N unexpected fields present). (L408 P2) The upstream prefilter SessionAuditRecords whitespace-splits the raw line BEFORE %q decoding, so session B's record whose free-form argv contains the substring session_id=A slips through when exporting A. auditEvent now re-checks the DECODED session_id against the target and SKIPS (emit=false, via a new (Event,bool,error) return) any record whose real session_id is absent or different — the authoritative match is the decoded key, so a spoofed record is never attributed to this session's app/device. Load-bearing: secret-in-key → no hunter2/--password in any name or value; session-B record with argv session_id=A → not emitted for A, genuine A still emitted. Exact CI validators zero (shfmt/shellcheck/codespell); go test -race green.

omkhar:
@codex review

PR #470, opened 2026-07-09
^F A5 (1/3): ocsf strict audit-line decode primitives (bare-token + dup-key reject, decoded session_id claim)

omkhar:
First of the A5 signed-session-audit stack (split from #466 to fit the 1200-line reviewer cap). The `internal/ocsf` strict-decode primitives the auditseal library (PR 2/3) consumes:
- **DecodeAuditLineStrict** — rejects bare (non key=value) tokens AND duplicate keys, so a tampered line that appends a bare token (e.g. `FORGED`) can't rebuild an identical digest (closes the P1 tamper hole from the original review).
- **AuditLineClaimsSession** — decodes then matches the session_id FIELD, so a spoofed `session_id=<id>` planted in another record's free-form argv is not mis-attributed (the F1 L408 class), and a malformed line that structurally claims a session is distinguishable from an unrelated one.
Both unit-tested (argv-spoof-safe + dup-key, both bash-%q and apple-container encodings). Honest layering: audit-line decoders live with the OCSF on-disk format; the crypto/chain library consumes them. 158 lines. dead-code clean (test callers reachable), EXACT CI validators zero.

### DIFF
--- a/internal/ocsf/auditline.go
+++ b/internal/ocsf/auditline.go
@@ -289,7 +289,12 @@ func splitQuotedTokens(line string) ([]string, error) {
 		case r == '$' && i+1 < len(runes) && runes[i+1] == '\'':
 			decoded, next, err := decodeAnsiC(runes, i+2)
 			if err != nil {
-				return nil, err
+				// Return the tokens completed BEFORE the malformed field so a
+				// caller that only needs the delimited fields parsed so far (e.g.
+				// AuditLineClaimsSession checking for a session_id field token that
+				// preceded the corruption) can act on them. decodeAuditLine still
+				// treats any error as fail-closed and discards these.
+				return tokens, err
 			}
 			cur.WriteString(decoded)
 			inToken = true
--- a/internal/ocsf/decode.go
+++ b/internal/ocsf/decode.go
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+package ocsf
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/omkhar/workcell/internal/applecontainer"
+)
+
+// AuditField is one ordered key/value pair decoded from an audit line. It is
+// the exported view of the package-internal auditField so that the A5 audit
+// hash-chain verifier (internal/host/auditseal) can reuse this package's
+// hardened, dup-key-rejecting tokenizer instead of reimplementing the bash
+// `printf %q` and percent-path decoders.
+type AuditField struct {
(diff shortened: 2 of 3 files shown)


omkhar:
@codex review

omkhar:
Fixed L83 in 3227d16. AuditLineClaimsSession returned false whenever splitQuotedTokens errored on a malformed quoted field, so a corrupted record for the TARGET session (session_id=sess-A note=$'unterminated) was misclassified unrelated instead of reaching DecodeAuditLineStrict to fail closed. Fix: splitQuotedTokens now returns the fields it completed BEFORE the error (backward-compatible — decodeAuditLine still fails closed + discards them; the unterminated-block test only asserts err!=nil), and AuditLineClaimsSession scans those, so a session_id FIELD token matching the target that preceded the corruption still claims. No L280 regression: an argv-embedded session_id stays one argv token (never a session_id field), and there are no completed field tokens past the error point, so it doesn't claim. Tests: matching field before a malformed field → claims (pre-fix false); argv-spoof with a malformed tail → still non-member for the spoofed id; the auditseal L254 test still passes, now backed by the corrected primitive. dead-code clean, validators zero.

omkhar:
@codex review

omkhar:
Fixed L90 in 837b640 (+ a5a2 verifier gate) — a clean architectural separator, not a tradeoff. The a5a2 verifier scopes chain verification to 0..head (L209), so a malformed line INSIDE the sealed range already fails closed; the real L90 gap is a malformed record appended AFTER the head whose session_id is unreadable (writer puts event= before session_id=). Key invariant: the genuine writer NEVER emits an untokenizable line (it always closes bash ANSI-C $'…' blocks). So: UNTOKENIZABLE (splitQuotedTokens error) = corruption/tamper → FAIL CLOSED (catches L90 regardless of field order, no attribution needed); TOKENIZABLE = attributable → claim-based membership (preserves L209 — a legitimate torn/legacy other-session line tokenizes cleanly + is attributed by its session_id; L280/L254 preserved). Verified against the actual L209 test (its torn line session_id=sess-B event=legacy_no_digest tokenizes fine → stays a skipped non-member). Added ocsf.AuditLineTokenizable + a Pass-1 corruption gate in auditseal. Load-bearing: L90 (malformed event=$'unterminated before session_id, appended same-session after head → verify FAILS) + L209 companion (unrelated tokenizable session_id=sess-B after head → this session PASSES). Accepted residual (crash-truncation inside an ANSI-C block fails readers of that log = correct fail-closed) documented in the trust model. dead-code clean, validators zero.

omkhar:
@codex review

PR #471, opened 2026-07-09
^F A5 (2/3): host-side audit-seal library — hash-chain verify + per-host signing

omkhar:
Second of the A5 stack (stacks on #470; base retargets to main once PR0 merges). The `internal/host/auditseal` crypto/chain library (Sigstore-independent — local per-host key, maintainer-approved trust model). Recomputes the existing audit hash-chain to the head, per-host ECDSA P-256 key (atomic publish, pubkey validate/repair, TOCTOU perms recheck), signs+verifies over the RECOMPUTED head, seal I/O. Consumes PR0's strict decode. Full tamper-matrix: bare-token / reorder / drop / dup-key / wrong-key / rotation / perms / no-chain (ErrUnsupportedAuditChain) / decoded-membership-spoof (L280) / malformed-same-session (L254) / key-store TOCTOU (L446). 1142 lines, dead-code clean, validators zero. Draft until PR0 merges (main-base rule).

### DIFF
--- a/internal/host/auditseal/auditseal.go
+++ b/internal/host/auditseal/auditseal.go
@@ -1,0 +1,266 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+// Package auditseal implements A5 "Signed Session Audit Records": it signs the
+// head of scripts/workcell's existing audit hash-chain host-side and verifies the
+// signature over the RECOMPUTED head from the authoritative durable log, so any
+// tamper fails closed. This is a boundary/host signature (not agent-signed): it
+// detects tamper by any party lacking the per-host key, but does NOT defend a
+// host-root attacker who can read signing.key. See
+// docs/signed-session-audit-records.md.
+package auditseal
+
+import (
+	"crypto/ecdsa"
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/omkhar/workcell/internal/host/hoststate"
+	"github.com/omkhar/workcell/internal/host/keystore"
+	"github.com/omkhar/workcell/internal/ocsf"
+)
+
+const Algorithm = "ecdsa-p256-sha256"
+
+const sealVersion = 1
+
+// ErrUnsupportedAuditChain reports that a session's audit records carry no digest
+// chain (the preview-only apple-container target). Callers treat it as
+// "unsigned": the hook skips it and `session verify` fails closed.
+var ErrUnsupportedAuditChain = errors.New("auditseal: session audit records have no digest chain (provider audit chain unsupported)")
+
+// Seal is the durable, host-owned signature over a session's audit-chain head,
+// stored beside the session record. Only Version, SessionID, KeyID, Algorithm,
+// and Signature are load-bearing; HeadDigest and SignedAt are informational
+// (verification always recomputes the head from the authoritative log).
+type Seal struct {
+	Version    int    `json:"version"`
+	SessionID  string `json:"session_id"`
+	HeadDigest string `json:"head_digest"`
+	KeyID      string `json:"key_id"`
+	Algorithm  string `json:"algorithm"`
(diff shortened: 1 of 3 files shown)


omkhar:
Rebased onto main (PR 1/3 a5d ocsf decode primitives merged); base retargeted main, marked ready. This is the auditseal crypto/chain library consuming the merged decode primitives. @codex review

omkhar:
Fixed L298 + L389 in b64e89c — extends the L446 key-store perms discipline to the EXISTING-key paths. (L298) adoptSigningKey (both direct-read and O_EXCL-loser) now runs requireOwnerSecure(signing.key, 0o077) BEFORE parsing — a signing.key that is a symlink, wrong-owner, or group/world-accessible is REFUSED (fail closed), not adopted. A private key is never repaired (a possibly-substituted key is rejected, not fixed). (L389) ensurePublicKey no longer early-returns on correct-content when the file is insecurely stored: it also checks requireOwnerSecure(path, 0o022); if the mode drifted (0666) or it's a symlink, it falls through to the atomic rewrite (temp+rename from the private key's public half — os.Rename replaces a symlink with a proper owner-owned regular file), so the .pub always passes the verify-side check and SignSessionHead no longer produces a self-inflicted seal that VerifySessionSeal rejects. A public key is repaired safely (derived from the 0600 private key we hold); the private key is refused. Tests: existing signing.key at 0644 / symlink → sign FAILS closed; correct-content .pub at 0666 / symlink → rewritten owner-owned + verify PASSES. All prior tests green; trimmed comment verbosity to 1199 (under cap). @codex review

omkhar:
Fixed L269 + L347 in 1003f2f — check-before-operate ordering. (L269) loadOrCreateSigningKey no longer os.ReadFiles an existing key before the security check: it Lstats + delegates to adoptSigningKey, which runs requireOwnerSecure(keyPath, 0o077) (lstat-based) BEFORE opening the file, in both the direct-load and O_EXCL race-loser branches. A signing.key that is a symlink (FIFO/device/huge file), wrong-owner, or group/world-accessible is refused WITHOUT ever being opened — signing can't block or read through a substituted path. (L347) ensureSecureDir now Lstats the directory first: a pre-existing symlink (or non-directory) is refused BEFORE any os.Chmod, so chmod/stat never operate on a symlink target + keys are never written through an attacker path; it creates the dir privately (0700) only when absent, then ends with requireOwnerSecure(dir, 0o077) as the final gate (catches a post-chmod swap). Tests: signing.key symlinked to a FIFO → sign fails closed with the symlink reason WITHOUT opening it (opening a FIFO would block forever, so the prompt error proves non-open); symlinked signing dir → fails before chmod, no signing.key written through the symlink. All prior tests green. @codex review

omkhar:
@codex review

omkhar:
Addressed in 57e0e7d: Pass 2 now skips the contiguous leading run of pre-chain (legacy) lines and anchors strict verification at the first digest-bearing line; a missing record_digest AFTER the chain root is still tamper. Tests cover legacy-prefix-verifies, mid-chain-strip-fails, and pure-legacy→ErrUnsupportedAuditChain. (Append serialization for the concurrent-fork finding lands with the session-verify command in the stacked PR, alongside its manifest regen, so the strict verifier and serialized appends ship together.)

omkhar:
@codex review

omkhar:
Addressed in a672a89: WriteSeal now fsyncs filepath.Dir(path) after the rename (both first-write and replace) and fails closed on any dir open/fsync error, so a seal that can't be made durable never reports success. Tests assert the dir fsync runs once per write + injected EIO fails WriteSeal closed on both paths.

omkhar:
@codex review

omkhar:
Addressed in 2d10d01: the no-chain guard now returns ErrUnsupportedAuditChain only when NO record in the chain region carries a record_digest (genuine no-chain provider); a stripped head digest or appended no-digest head on a chained provider now falls through to Pass 2 as chain-broken tamper (HasSignableChain stays true). Tests assert stripped-head and appended-no-digest-head are tamper-not-unsupported.

omkhar:
@codex review

PR #472, opened 2026-07-09
^F A5 (3/3): workcell session verify + session-stop signing hook + trust-model doc

omkhar:
Third of the A5 stack (stacks on PR 2/3; base retargets to main as parents merge). Wires the auditseal library into the product: `workcell session verify --id` (recompute chain over the authoritative durable log, dup-key reject, verify head signature over the recomputed head, fail-closed 0/1/2, G2-redacted); the session-stop signing hook (finalize_session_audit); L240 audit-append flock serialization (via the atomic acquire-profile-lock primitive — no flock(1), absent on macOS); full dispatch/usage/man/operator-contract/manifest lockstep; trust-model doc (boundary/host-signed, detects offline/exported/reorder/drop/dup-key/malformed tampering, NOT host-root re-sign; apple-container unsigned fail-closed follow-up + the audit-append serialization invariant). 776 lines, validators zero. Draft until parents merge.

### DIFF
--- a/cmd/workcell-hostutil/main.go
+++ b/cmd/workcell-hostutil/main.go
@@ -99,6 +99,10 @@ func run(args []string) error {
 		return cmdHelperSessionStopCli(args[1:])
 	case "session-timeline-cli":
 		return cmdHelperSessionTimelineCli(args[1:])
+	case "session-verify-cli":
+		return cmdHelperSessionVerifyCli(args[1:])
+	case "session-sign-head":
+		return cmdHelperSessionSignHead(args[1:])
 	case "support-bundle-cli":
 		return cmdHelperSupportBundleCli(args[1:])
 	case "support-bundle-usage":
@@ -370,6 +374,14 @@ func cmdHelperSessionTimelineCli(args []string) error {
 
 func cmdHelperSessionLogsCli(args []string) error {
 	return sessionctl.LogsMain(args)
+}
+
+func cmdHelperSessionVerifyCli(args []string) error {
+	return sessionctl.VerifyMain(args)
+}
+
+func cmdHelperSessionSignHead(args []string) error {
+	return sessionctl.SignHeadMain(args)
 }
 
 func cmdHelperSessionAttachCli(args []string) error {
@@ -1013,7 +1025,7 @@ func parsePrepareBundleArgs(args []string) (*injection.PrepareBundleOptions, err
 // already do); previously these returned plain errors and collapsed to the
 // exit-1 fallback, an intra-binary inconsistency (D8).
 func usage() error {
-	return &cliexit.ExitCodeError{Code: 2, Message: "usage: workcell-hostutil <path|release|helper|launcher|policy|resolve-credentials|pty-transcript|auth-cli|auth-usage|policy-cli|policy-usage|publish-pr-cli|publish-pr-usage|session-usage|session-attach-cli|session-delete-cli|session-dispatch-cli|session-logs-cli|session-monitor-cli|session-send-cli|session-stop-cli|session-timeline-cli|support-bundle-cli|support-bundle-usage> [args...]"}
+	return &cliexit.ExitCodeError{Code: 2, Message: "usage: workcell-hostutil <path|release|helper|launcher|policy|resolve-credentials|pty-transcript|auth-cli|auth-usage|policy-cli|policy-usage|publish-pr-cli|publish-pr-usage|session-usage|session-attach-cli|session-delete-cli|session-dispatch-cli|session-logs-cli|session-monitor-cli|session-send-cli|session-stop-cli|session-timeline-cli|session-verify-cli|session-sign-head|support-bundle-cli|support-bundle-usage> [args...]"}
 }
 
 func pathUsage() error {
--- a/docs/enterprise-evidence-baseline.md
+++ b/docs/enterprise-evidence-baseline.md
@@ -29,6 +29,11 @@ Current audit and session evidence is host-local:
 - audit records are append-only text records with timestamps, event fields,
   assurance data, and chained record digests
 - launched sessions write durable JSON records under the same target-state root
+- session audit-chain heads are signed host-side, and
+  `workcell session verify` recomputes the chain from the authoritative log and
+  checks that signature — see
+  [signed-session-audit-records.md](signed-session-audit-records.md) for the
+  format and trust model (boundary/host-signed, not agent-signed)
 - `workcell session timeline` filters audit records for one session
 - `workcell session export` bundles a session record with matching audit records
 - durable session records intentionally survive `workcell --gc`
--- a/docs/incident-response.md
+++ b/docs/incident-response.md
@@ -258,12 +258,26 @@ you can detect truncation or replacement of the durable log against a preserved
 snapshot (step 3). Cross-check the exported records, the timeline, and the
 preserved `workcell.audit.log` for consistency.
 
-**What does not exist yet.** There is **no cryptographic hash-chain
-verification** of session audit records today. Signed, tamper-evident audit
-records with external verification tooling are **roadmapped as A5**
-(["Signed Session Audit Records", milestone v0.15](../ROADMAP.md#track-a-boundary-depth-and-agent-threat-defenses)) —
-a `workcell session verify`-style command is planned there but is **not
-shipped**. Until A5 lands, integrity is host-preservation plus consistency
+**Cryptographic verification (A5).** Session audit records now form a signed,
+tamper-evident hash chain, and `workcell session verify --id SESSION_ID`
+recomputes that chain from the **authoritative** durable profile log and verifies
+the host-side signature over the chain head against the per-host signing key
+(["Signed Session Audit Records"](../ROADMAP.md#track-a-boundary-depth-and-agent-threat-defenses)).
+As a responder, run it against the suspect session: a `session_verify=verified`
+result means the audit chain is intact and its head is signed by this host's key,
+so the recorded events have not been altered offline. It fails closed (non-zero,
+no `verified`) on any tamper the chain detects — a flipped byte, a reordered,
+dropped, or duplicated record, an appended container-side line, an
+exported/copied log substituted for the real one, or a missing/mismatched seal.
+It does **not** defend against an attacker with host root who re-signs a rewritten
+chain with the host key (out of scope per the A5 threat model), so pair a
+`verified` result with the host-preservation and consistency cross-checks above
+rather than treating it as proof the host itself was not compromised.
+
+**What still relies on preservation.** For providers whose audit records carry no
+digest chain (e.g. the preview-only apple-container target), `session verify`
+fails closed as unsupported rather than asserting integrity; there, and for the
+host-root case above, integrity is host-preservation plus consistency
 cross-checks, not signature verification. Do not represent current audit records
 as cryptographically tamper-proof.
 
--- a/docs/safe-path-expectations.md
+++ b/docs/safe-path-expectations.md
@@ -62,6 +62,7 @@ workcell session logs --id 20260408T120000Z-1a2b3c4d --kind audit
 workcell session timeline --id 20260408T120000Z-1a2b3c4d
 workcell session diff --id 20260408T120000Z-1a2b3c4d
 workcell session export --id 20260408T120000Z-1a2b3c4d --output /tmp/workcell-session.json
+workcell session verify --id 20260408T120000Z-1a2b3c4d
 workcell policy show
 workcell policy diff
 workcell why --agent codex --mode strict --credential codex_auth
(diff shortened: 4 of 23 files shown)


omkhar:
Rebased onto main (a5a2 auditseal-lib now merged) → 85ce8e48. This is the final A5 PR: the `workcell session verify` command + the audit-append serialization (dedicated per-log lock around the read-digest→append critical section in scripts/workcell, with control-plane-manifest regen) that makes the strict linear chain the verifier enforces actually hold under concurrent writers.

omkhar:
@codex review

omkhar:
Addressed in 0fb75a2: the stop-fallback terminal path now signs too — extracted sign_session_audit_head_explicit(profile,session_id,record_path), wired into all three fallback terminal-record writes (the failed-case write gated on !session_monitor_pid_is_live so a live monitor still owns signing → no double-sign). Fail-closed on no-chain. Manifest regenerated. E2E: killed-monitor stop-fallback session now gets .audit-sig and session verify=verified.

omkhar:
@codex review

omkhar:
Addressed in 323f9d6 (support_bundle_main pattern, defense in depth): (P1) session_verify_main now REJECTS user-supplied --signing-dir/--real-home (both = and bare forms) with exit 2 before the helper runs, AND appends the host-owned trusted flags AFTER user args so trusted-wins even on a missed injection; test asserts a user --signing-dir=<evil> is rejected and never prints session_verify=verified while genuine verify still works. (P3) session delete now removes the <id>.audit-sig sidecar (in plan, dry-run would_remove, and rm), kept under --record-only. Manifest regenerated.

omkhar:
@codex review

omkhar:
Addressed in 749cad5: systematic audit of all delete-expectation assertions. Only L2822 (cli-life, real CLI — session is sealed via the verify path, confirmed by the round-8 e2e seal-exists check) needed audit_seal added to removed=; verified byte-for-byte against session_delete_main's plan ordering (audit_seal appended last). L2466 (detached-lifecycle mock) empirically confirmed NOT sealed at delete (SEAL_EXISTS_BEFORE_DELETE=0 — the mock's audit log resolves outside COLIMA_STATE_ROOT so signing fail-closes to a no-op), tampered/unsigned assertions verified unchanged. Rebased onto current main.

omkhar:
@codex review

omkhar:
Addressed in 62ae0c0 (defense in depth): verify now derives the audit log path CANONICALLY from the trusted on-disk record location (grandparent of recordPath = profile state dir, matching hoststate.ProfileAuditLogPath/LegacyProfileAuditLogPath) instead of the free-form record.AuditLogPath, AND fail-closes (exit 1) if the recorded path != canonical. Repro test: pristine-copy-elsewhere + tampered-canonical-log + record-repointed → verify does NOT report verified. Also corrected verify fixtures that had the audit log inside sessions/ (masking the bug) to the real sibling layout. Genuine/tamper/unsigned/no-chain/legacy-prefix tests all green.

omkhar:
@codex review

omkhar:
Addressed in 33faec9: the signer now derives its audit log the SAME grandparent-of-record way verify does — `$(dirname $(dirname record_path))/workcell.audit.log` — so signer path == verifier canonical path is provably identical for modern AND legacy layouts (modern still equals profile_audit_log_path, so launch finalize is unchanged). Legacy scenario test asserts signer_matches_legacy_log=1, seal_exists=1, verified=1; negative control confirms the old profile_audit_log_path signer skipped the seal on legacy. Manifest regenerated.

omkhar:
@codex review

omkhar:
Addressed in 5cd3a0f: raised the audit-lock wait budget 50→1500 attempts (~30s sleep) so a waiter reliably outlasts a cold `go run` digest computation by the lock holder, instead of exhausting a 1s budget and skipping a terminal audit/signing record. Fail-closed on genuine exhaustion preserved (now a real stuck-holder anomaly). Digest still computed under the lock (no fork). Comment + manifest updated. The slow-holder regression test is in stacked test-only PR (over the 1200-line cap here).

omkhar:
@codex review

omkhar:
Addressed in f7370a9: session_verify_main now REJECTS a user-supplied --root (both --root=VALUE and bare --root VALUE, exit 2) like --signing-dir/--real-home, so a user can't inject an alternate lookup root and verify a host-signed COPY under it; legitimate host-config env (WORKCELL_STATE_ROOT/COLIMA_STATE_ROOT) that derives the trusted root stays honored. Systematic sweep of ALL verify inputs confirms --root was the last CLI redirect vector — record.AuditLogPath (canonicalized), TargetProvider (decoder-only, fail-closed), identity fields (safe by construction), env vars (host config, same trust level as state). Test in stacked a5b2t. Manifest regenerated.

omkhar:
@codex review

omkhar:
Fixed 3 deterministic session-commands scenario test bugs that the Validate-repository job surfaced (reproduced + fixed locally against real colima docker; full scenario now passes end-to-end, EXIT=0). All three are TEST-harness defects — the product sign/verify/delete code is correct (the trace shows session_verify=verified succeeding):

1. **Legacy sign/verify block**: `go_hostutil session-verify-cli … | grep -q` under `set -o pipefail` was racy — `grep -q` exits on the first matching line and closes the pipe, so the verifier takes SIGPIPE writing its remaining lines and exits non-zero, which pipefail surfaces as a failed pipeline even though verification succeeded. Fixed: capture output first, then grep (matches the pattern the other verify assertions already use).
2. **Detached-lifecycle delete**: round-11's grandparent-of-record signer change made that mocked session get sealed at stop, so `removed=` now includes `audit_seal`; updated the stale expectation.
3. **cli-life block**: `EVIL_SIGNING_DIR="${TMP_DIR}/…"` referenced ${TMP_DIR}, which isn't in the inner `bash -lc`'s scope → `set -u` unbound-variable fatal. Fixed to ${PROFILE_DIR} (in-scope, cleaned up by the block).

shellcheck -x / shfmt -ln=bash -i 2 -ci / codespell all clean on the file. The audit-lock-budget + verify-hardening test PR is rebased onto this head.

@codex review

omkhar:
@codex review

omkhar:
Addressed in 5eaf185: the signer now takes the SAME <audit-log>.lock append_audit_record_to_path uses around its head read (go_hostutil session-sign-head), so a concurrent appender's torn multi-printf record can't make SignSessionHead skip the seal. Release on every path; ~30s wait budget + stale reclaim; verified no nested-lock deadlock across all 3 signing call sites (append helpers release before the signer runs). Positive: signer blocks ~2.2s behind a held lock then signs a verified head; negative control: unlocked signer skips the seal on a torn read. Manifest regenerated. Validated against the REAL colima session-commands scenario (EXIT=0), which also caught + fixed a latent TMP_DIR-unbound in the prior --root test. Concurrency test in stacked a5b2t.

omkhar:
@codex review

omkhar:
Addressed in c5ed997: updated the stale FR-007 docs in lockstep — incident-response.md §5 rewritten (verify now SHIPS: recomputes the chain from the authoritative log + verifies the head signature against the per-host key; enumerates what it detects — flipped byte/reorder/drop/dup-key/appended-container-line/exported-log-substitution/mismatched-seal — and, without overclaiming, that it does NOT defend against host-root re-sign or no-chain providers), plus safe-path-expectations.md, workcell-system-design.md, and workcell-session-supervisor-design.md now list session verify. To stay under the 1200 cap I moved verify_test.go (294 lines) to the stacked test PR a5b2t and migrated its requirements.toml/operator-contract.toml evidence citations there too (verify stays backed by the e2e scenario + auditseal_test.go in a5b2); coverage + operator-contract green on both branches. a5b2 now 933 lines. check-dead-code clean.

omkhar:
@codex review

omkhar:
Addressed in 2b99ac0: closed the TOCTOU signing race — session_stop_main now, once the record is terminal, RECHECKS for a missing .audit-sig and signs it if absent, gating on seal-presence rather than monitor-liveness. So if the monitor dies in the window between the live-check and its own finalize signing, the stopped session is still sealed. Idempotent ([[ ! -e …audit-sig ]] guard → no double-sign, re-run no-op); additive to the round-8 monitor-dead fallback; uses the grandparent-of-record derivation + audit-log lock. Evidence: race→seal=1/verified=1; pre-signed→unchanged=1; negative control (recheck removed)→seal=0. Real colima scenario EXIT=0. Manifest regenerated.

omkhar:
@codex review

omkhar:
Addressed in 7f17c88 (audit-log lock L2625 only; the session/profile lock at L2478 intentionally stays $$ since its lifetime is the top-level session): records the holder as `${BASHPID:-$$}` instead of `$$`.

Note on the portable form (not bare `${BASHPID}`): scripts/workcell's shebang is `/bin/bash`, which on the macOS host is **bash 3.2.57 — no BASHPID**; bare ${BASHPID} errors 'unbound variable' under set -u (the required real-colima scenario run caught this on the first attempt). `${BASHPID:-$$}` uses the subshell's real PID wherever BASHPID exists (bash 4+, incl. the Linux container → fixes subshell stale-reclaim) and safely degrades to $$ on bash 3.2 (no regression vs. before). Fully fixing subshell reclaim on the bash-3.2 host has no portable subshell-PID primitive and would need a shebang bump to bash 4+ — a larger, separate change, out of scope here.

Evidence (harness bash 5.3): backgrounded subshell holder kill -9'd before cleanup → next append reclaims + appends (chain linear, waited=0); negative control ($$) hangs on the budget. check-dead-code clean, real scenario EXIT=0, manifest regenerated. Reclaim test in stacked a5b2t.

omkhar:
@codex review

omkhar:
Addressed in 1a2faaa:

**L107 (verify output contract):** added `session_verify=`, `key_id=`, `head_digest=` to `policy/public-contract.toml` `[output_lines].prefixes` and `docs/stability-contract.md`'s stable-output table, so the public-contract validator now guards verify's output against renames/removals. check-public-contract passes.

**L2637 (audit-lock reclaim) — correctness fix + scoped push-back:** while addressing this I found the prior `${BASHPID:-$$}` was evaluated INSIDE the `$(go_hostutil …)` command substitution, so on bash 4+ it captured that transient subst-subshell's PID (already dead by the stale check) — subtly wrong. Fixed by HOISTING the capture into the holder's own context (`local holder_pid="${BASHPID:-$$}"` before the loop), verified: in a backgrounded subshell holder_pid == the subshell's real PID (not the transient $() one), and == $$ on bash 3.2.

On the reclaimable-on-bash-3.2 concern specifically: I traced every audit-lock holder — append_audit_record_to_path callers, sign_session_audit_head_explicit, finalize_session_audit, and the detached `session monitor` (which runs as its OWN top-level /bin/bash process). On the macOS host path every holder is a **top-level process** (launch shell, monitor process, session-stop/-control command), whose `$$` IS the real holder — so reclaim works on bash 3.2. A same-process backgrounded-subshell holder only arises under bash 4+ (the concurrency tests), where the hoisted BASHPID now records it correctly. bash 3.2 has no portable primitive for a subshell's own PID; since no such host holder exists, $$ is correct there and no age-ceiling is needed. Real colima scenario EXIT=0; manifest regenerated.

@codex review

omkhar:
@codex review

omkhar:
Addressed in ad4bc3a: (L61) the FindSessionRecordWithPathInRoots error branch no longer returns the raw lower-layer error (which embeds the absolute record path) — it now returns a support-bundle-redacted ExitCodeError (`redactor.String(...)`, home→~, credentials masked) like every other message in verifyMain, so verify honors its redacted-output contract even on the damaged-record paths operators inspect. (L52) corrected the trust-model doc's audit-lock wait budget from ~1s to the actual ~30s (1500×20ms, to outlast a cold go-digest append), matching the code + comment. go build/vet/test + dead-code + markdownlint + doc-links clean; real colima scenario EXIT=0.

omkhar:
@codex review

omkhar:
CI fix in a3fcf43: the L107 addition of session_verify=/key_id=/head_digest= to public-contract.toml [output_lines].prefixes broke internal/metadatautil's TestCheckPublicContractRejectsBogusOutputPrefix — its mutation anchor hardcoded `"egress_enforcement="]` (assuming egress was the LAST prefix), which no longer matches now head_digest= is last. Updated the anchor to `"head_digest="]`. Full `go test ./... -race` now green (0 FAIL). (Root cause of the miss: I validated with check-public-contract.sh + the scenario but not the full go-test suite; corrected.)

omkhar:
@codex review

omkhar:
Addressed in d931cc2: the already-stopped + live-monitor early-return branch of session_stop_main (container stopped, monitor observed live) returned without reaching the round-13 seal-missing recheck below it. If the monitor wrote the terminal record but exited before its finalize_session_audit signing step, that session was left unsigned. Added the SAME idempotent recheck before that branch's return 0 — gated on a terminal record AND an absent .audit-sig (a monitor that already signed is not double-signed). Mirrors the tested fall-through recheck. shellcheck/shfmt clean, full go test ./... 0 FAIL, real colima scenario EXIT=0, manifest regenerated.

omkhar:
@codex review

PR #478, opened 2026-07-09
A5 PR-A: keystore — per-host signing key management for audit seals

omkhar:
Part of A5 (signed session audit seals). First PR in a 4-PR stack.

Manages the per-host ECDSA **P-256** key used to sign A5 session audit seals: generate, load, and harden the private key and its public sidecar under an owner-secured (0700) signing directory. Distinct concern from the audit-chain recompute + sign/verify in package `auditseal` (stacked PR-B), which consumes `LoadOrCreateSigningKey`/`LoadPublicKey`.

**Containment.** The whole operation is anchored to one validated directory fd: the signing dir is opened `O_NOFOLLOW|O_DIRECTORY` and fstat-validated (real dir, owner euid, 0700 via `unix.Fstat`), then every key/pub read and write is an `*at` syscall relative to that fd (`openat`/`renameat`/`linkat`/`unlinkat`), so a post-validation dir swap cannot redirect any read or write. Content files are `O_NOFOLLOW` + fstat-checked (regular file, owner euid, mode). Keys are **refused** (never repaired) if insecure/symlink/non-regular/wrong-owner/wrong-curve; the public sidecar is repaired atomically from the private key and bound to its key-id filename. Directory entries are made durable by fsyncing the dir and **its full ancestor chain**, independent of who won a first-use race.

**Core tests** here: key gen + stability + rotation, P-256 curve enforcement, one perms refusal, one durability (parent fsync) case, blank-dir guard. Dead-code clean standalone. The exhaustive hardening matrix (perms/symlink/FIFO/TOCTOU/durability/racers) is stacked test-only PR-A2.

Stack:
- **PR-A (this)** a5k1/keystore-core → main
- PR-A2 a5k2/keystore-hardening-tests → a5k1 (test-only)
- PR-B a5a2/auditseal-lib → a5k1
- PR-C a5b2/session-verify → a5a2

### DIFF
--- a/internal/host/keystore/keystore.go
+++ b/internal/host/keystore/keystore.go
@@ -1,0 +1,566 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+// Package keystore manages the per-host ECDSA P-256 key used to sign A5 session
+// audit seals: generating, loading, and hardening the private key and its
+// public sidecar under an owner-secured signing directory. It is a distinct
+// concern from the audit-chain recompute and sign/verify in package auditseal,
+// which consumes LoadOrCreateSigningKey and LoadPublicKey.
+//
+// The ENTIRE operation is anchored to one validated directory fd. The signing
+// dir is opened O_NOFOLLOW|O_DIRECTORY and fstat-validated (real dir, owner euid,
+// 0700), then every key/pub read and write is performed with *at syscalls
+// relative to that fd (openat/renameat/linkat/unlinkat). So a same-parent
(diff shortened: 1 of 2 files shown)


omkhar:
@codex review

omkhar:
Addressed in af7a5d1 (rebased to fede5861): the ancestor-durability walk is now fd-anchored (openat(fd, "..") from the validated dir fd, dev+inode root fixpoint) instead of path-based, so a symlinked ancestor can no longer short-circuit the fsync of the real directory behind it. Added TestSymlinkedAncestorIsFsynced (a5k2) asserting the real dir's inode is fsynced. Rebased onto current main (post-g4).

omkhar:
@codex review

PR #480, opened 2026-07-09
A5 PR-C2: audit-lock wait-budget slow-holder test (test-only)

omkhar:
Stacked on **a5b2/session-verify** (#472). **Test-only** — the production fix (audit-lock wait budget 50→1500 to outlast a cold `go run` digest call) lives in a5b2; this carries the slow-holder scenario test that pushed a5b2 over the 1200-line PR-shape cap. Stays draft until a5b2 merges, then rebases onto main.

The test stubs the lock-acquire attempts instant (pure-bash `mkdir`, leaving the real `audit-digest` intact) so the wait budget is exactly `cap × 20ms` of pure sleep, holds the lock ~2s, and asserts the second writer still appends (`waiter_rc=0`, `lines=2`) and chains linearly. Deterministic negative control: cap=50 → 1s budget → "skipped a record" (bug); cap=1500 → 30s → both records land linear.

### DIFF
--- a/internal/sessionctl/verify_test.go
+++ b/internal/sessionctl/verify_test.go
@@ -1,0 +1,294 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+package sessionctl
+
+import (
+	"bytes"
+	"errors"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/omkhar/workcell/internal/cliexit"
+	"github.com/omkhar/workcell/internal/host/hoststate"
+	"github.com/omkhar/workcell/internal/host/sessions"
+)
+
+// verifyFixture lays a genuine, signed session on disk and returns the state
+// root, signing dir, session id, and the durable audit-log path so a test can
+// tamper it. The record uses the legacy `<root>/<profile>/sessions` layout that
+// sessions discovery accepts.
+func verifyFixture(t *testing.T) (root, signingDir, sessionID, logPath string) {
+	t.Helper()
+	root = t.TempDir()
+	signingDir = filepath.Join(t.TempDir(), "signing")
+	sessionID = "session-1"
+
+	// Canonical layout: the audit log is a SIBLING of the sessions dir under the
+	// profile state dir (<root>/<profile>/workcell.audit.log), exactly where
+	// profile_audit_log_path/the signer place it and where verify recomputes it.
+	logPath = filepath.Join(root, "wcl-fixture", "workcell.audit.log")
+	if err := os.MkdirAll(filepath.Join(root, "wcl-fixture", "sessions"), 0o700); err != nil {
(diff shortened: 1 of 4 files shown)


omkhar:
Rebased onto main (a5b2 now merged) → 2f6870da. This is the A5 test-overflow PR: re-adds internal/sessionctl/verify_test.go (294 lines) with its requirements.toml/operator-contract.toml evidence citations, plus the audit-lock-budget / signer-lock / subshell-reclaim / --root-reject / stop-recheck scenario test blocks that pushed a5b2 over the 1200-line PR-shape cap. Full go test ./... 0 FAIL, real colima session-commands scenario EXIT=0, shellcheck/shfmt clean, check-pr-shape 457 lines.

omkhar:
@codex review

omkhar:
Addressed in 264311b2: replaced the fixed `sleep 0.3` with a bounded poll for the holder's lock dir (`${audit_path}.lock`) before starting the waiter. append_audit_record_to_path takes the lock BEFORE the injected audit_record_digest sleep, so polling the lock dir deterministically guarantees the holder is holding it when the waiter starts — the slow-holder regression is now exercised regardless of CI scheduling latency (a delayed holder can no longer lose the race and pass spuriously). ~10s bound fails closed. shellcheck/shfmt clean, real colima scenario EXIT=0.

omkhar:
@codex review

omkhar:
Addressed in 31ee60d: (L305) restored tests/scenarios/shared/test-session-commands.sh to the [workflows.session_verify] evidence alongside the new unit tests — it's the only evidence exercising the public scripts/workcell wrapper (the --root/--signing-dir rejection is wrapper-only). (L962) replaced the elapsed-time proof-of-blocking (end-start>=1, which a cold go-sign satisfies even if the signer stopped locking) with an ORDERING assertion: hold the lock with a live PID, run the signer in the background, and assert NO seal appears while the lock is held (an ordering fact that can only flip after release), then release + confirm seal+verify. shellcheck/shfmt clean, verify-operator-contract 0, full go test ./... 0 FAIL, real colima scenario EXIT=0.

omkhar:
@codex review

omkhar:
Addressed in f85db4b3: (L965) instrumented the lock ACQUISITION instead of relying on signing elapsed time — the signer-lock test now mocks go_hostutil acquire-profile-lock to record each attempt (real mkdir semantics; session-sign-head + session-verify-cli stay real) and marks the signer blocked only once it has CONTENDED for the held lock >=2 times (marker >=3 bytes), deterministically, independent of cold-go-run startup; a regression that stops taking the lock records no attempt (times out) or seals immediately (caught). (L3187) the --root reject test now exercises BOTH the leading root-run position (--root before --id, the ConsumeRootArgs path the reject exists to close) AND trailing, for --root= and bare --root. shellcheck/shfmt clean, full go test ./... 0 FAIL, real colima scenario EXIT=0.

omkhar:
@codex review

PR #489, opened 2026-07-10
Refactor Codex adapter subcommand gate to deny-by-default allowlist

omkhar:
## Summary
- Refactor the Codex adapter subcommand gate from a blocklist to an **allowlist (deny-by-default)**. The historical blocklist reopened on every codex CLI version bump — a review of the 0.143 bump surfaced 15 real bypasses (new subcommands `plugin`/`remote-control`/`exec-server`/`mcp-server`/`update`, `features enable`, the `--yolo` alias, inline-TOML/quoted/escaped config-key evasions, `AGENT_UI=gui` env override). Deny-by-default closes the whole class: any new or unknown subcommand is denied until explicitly vetted.
- Permit set (classified read-only/session surface, verified against real `codex --help`): `exec/e, review, login, logout, completion, doctor, apply/a, resume, fork, archive, unarchive, delete, help, debug, execpolicy`; `features`/`features list` (read-only) with `features enable|disable` denied; `app`/`app-server` permitted only under `AGENT_UI=gui` (the managed GUI backend). Bare `codex`, `codex --version`, and `codex -- <prompt>` are permitted (default session/prompt path).
- Retains all config-override and dangerous-flag hardening for permitted subcommands (they still take `-c`/flags): `--yolo` + `--dangerously-bypass-*` glob, value-taking-global consumption, `-c` key normalization (quoted/single-quoted/whitespace/backslash-escape fail-closed), inline-table blocking (`features={…}`/`profiles={…}`/`mcp*`/`hooks`), and the `plugins/plugin_sharing/remote_plugin` feature pins across `.codex/config.toml`, `managed_config.toml`, `requirements.toml`.
- No upstream pin change in this PR — the codex version bump to the latest stable ships separately, riding on top of this allowlist so its new surface is auto-covered.

## Validation
- ./scripts/pre-merge.sh --profile pr-parity
- Sourced-function permit/deny harness 60/60; deny-by-default regression-locked (`codex frobnicate`/`some-new-daemon` denied; `codex -- frobnicate` permitted as prompt text).
- Real colima `scripts/container-smoke.sh` (SMOKE_EXIT=0) and `tests/scenarios/shared/test-session-commands.sh` (SCENARIO_EXIT=0) — the colima run empirically caught the hidden runtime `execpolicy` subcommand and added it to the permit set.
- Full go test ./..., shellcheck/shfmt clean, control-plane manifest regenerated + verified on the committed tree.

## Review
- Draft for the Codex review loop. Supersedes the blocklist hardening on the upstream-refresh PR.

### DIFF
--- a/adapters/codex/.codex/config.toml
+++ b/adapters/codex/.codex/config.toml
@@ -47,3 +47,10 @@ network_access = false
 # Codex's unified exec path currently fails inside Workcell managed mode with
 # "Failed to create unified exec process: Operation not permitted".
 unified_exec = false
+# Defense in depth for the 0.143+ plugin/marketplace surface: even though the
+# provider policy rejects the `plugin` subcommand outright, pin the plugin
+# feature flags off in the runtime baseline so no config layer can silently
+# reintroduce remote marketplace fetch/install or plugin sharing.
+plugins = false
+plugin_sharing = false
+remote_plugin = false
--- a/adapters/codex/README.md
+++ b/adapters/codex/README.md
@@ -55,16 +55,27 @@ In-container reserved session targets: `~/.codex/{config.toml,auth.json,`
   copy only in explicit lower-assurance cases (see
   [../../docs/adapter-control-planes.md](../../docs/adapter-control-planes.md#codex-rules-mutability)).
 - Unsafe-argument policy (`reject_unsafe_codex_args` in
-  `runtime/container/provider-policy.sh`): the wrapper blocks
-  `--dangerously-bypass-approvals-and-sandbox`, `--full-auto`, `-a`/
-  `--ask-for-approval`, `--add-dir`, `--search`, `--remote`, `--enable`/
-  `--disable`, `--cd`, `--sandbox danger-full-access`, reserved `--config`
-  overrides, off-mode `--profile` values, and `app`/`app-server`/`cloud`/`mcp`/
-  `sandbox` subcommands outside the managed GUI path. These are rejected in
-  **every** mode including `breakglass`: `provider-wrapper.sh` re-checks arguments
-  (`WORKCELL_WRAPPER_CONTEXT=1`), so `container-smoke.sh` confirms breakglass
-  overrides still fail. Breakglass raises the sandbox floor, not the unsafe-flag
-  policy.
+  `runtime/container/provider-policy.sh`): **subcommands are deny-by-default**
+  over the pinned Codex subcommand namespace — only the read-only/session
+  surface is permitted (exec, review, login/logout, completion, doctor, apply,
+  resume, fork, archive/unarchive/delete, help, debug, execpolicy, `features
+  list`); everything else is rejected (plugin, remote-control, exec-server,
+  mcp*, cloud*, responses-api-proxy, stdio-to-uds, update, sandbox, `features
+  enable`/`disable`, and any unclassified/new subcommand). `app-server` is the
+  managed GUI backend and is permitted only as the bare no-arg launch; an
+  unknown first token is Codex prompt text. The classified set is pinned to the
+  version by
+  [`tests/fixtures/codex-subcommands.txt`](../../tests/fixtures/codex-subcommands.txt)
+  (a `CODEX_VERSION` bump fails CI until it is regenerated). The wrapper also
+  blocks `--yolo`/`--dangerously-bypass-*`, autonomy/network flags
+  (`--full-auto`, `-a`, `--add-dir`, `--search`, `--remote`, `--enable`,
+  `--disable`, `--cd`, `--sandbox danger-full-access`), off-mode `--profile`
+  values, and reserved `-c`/`--config` overrides of guarded namespaces
+  (features/plugins/marketplaces/mcp/hooks/sandbox/approval, incl. glued,
+  quoted/escaped, inline-table, and `profiles.<name>.<key>` spellings). All are
+  rejected in **every** mode including `breakglass` (`provider-wrapper.sh`
+  re-checks with `WORKCELL_WRAPPER_CONTEXT=1`); breakglass raises the sandbox
+  floor, not the unsafe-argument policy.
 - Final branch publication stays on the host through `workcell publish-pr`, not
   from inside the container session.
 
--- a/adapters/codex/README.md
+++ b/adapters/codex/README.md
@@ -58,10 +58,10 @@ In-container reserved session targets: `~/.codex/{config.toml,auth.json,`
   `runtime/container/provider-policy.sh`): **subcommands are deny-by-default**
   over the pinned Codex subcommand namespace — only the read-only/session
   surface is permitted (exec, review, login/logout, completion, doctor, apply,
-  resume, fork, archive/unarchive/delete, help, debug, execpolicy, `features
+  resume, fork, archive/unarchive/delete, help, execpolicy, `features
   list`); everything else is rejected (plugin, remote-control, exec-server,
-  mcp*, cloud*, responses-api-proxy, stdio-to-uds, update, sandbox, `features
-  enable`/`disable`, and any unclassified/new subcommand). `app-server` is the
+  mcp*, cloud*, responses-api-proxy, stdio-to-uds, update, sandbox, debug,
+  `features enable`/`disable`, and any unclassified/new subcommand). `app-server` is the
   managed GUI backend and is permitted only as the bare no-arg launch; an
   unknown first token is Codex prompt text. The classified set is pinned to the
   version by
--- a/adapters/codex/README.md
+++ b/adapters/codex/README.md
@@ -68,8 +68,9 @@ In-container reserved session targets: `~/.codex/{config.toml,auth.json,`
   [`tests/fixtures/codex-subcommands.txt`](../../tests/fixtures/codex-subcommands.txt)
   (a `CODEX_VERSION` bump fails CI until it is regenerated). The wrapper also
   blocks `--yolo`/`--dangerously-bypass-*`, autonomy/network flags
-  (`--full-auto`, `-a`, `--add-dir`, `--search`, `--remote`, `--enable`,
-  `--disable`, `--cd`, `--sandbox danger-full-access`), off-mode `--profile`
+  (`--full-auto`, `-a`, `--add-dir`, `--search`, `--remote`,
+  `--remote-auth-token-env`, `--enable`, `--disable`, `--cd`,
+  `--sandbox danger-full-access`), off-mode `--profile`
   values, and reserved `-c`/`--config` overrides of guarded namespaces
   (features/plugins/marketplaces/mcp/hooks/sandbox/approval, incl. glued,
   quoted/escaped, inline-table, and `profiles.<name>.<key>` spellings). All are
--- a/adapters/codex/managed_config.toml
+++ b/adapters/codex/managed_config.toml
@@ -49,6 +49,13 @@ network_access = false
 # Codex's unified exec path currently fails inside Workcell managed mode with
 # "Failed to create unified exec process: Operation not permitted".
 unified_exec = false
+# Defense in depth for the 0.143+ plugin/marketplace surface: even though the
+# provider policy rejects the `plugin` subcommand outright, pin the plugin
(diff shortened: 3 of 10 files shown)


omkhar:
@codex review

omkhar:
Addressed in 53b4b3fa: (P2) deny-by-default is now scoped to the SUBCOMMAND NAMESPACE, not all first tokens — the gate classifies every pinned-codex subcommand as ALLOW / GUI-gated / DENY (exact-token), and a first bare token that is NOT a known subcommand is permitted as PROMPT text, so codex "fix tests" and workcell --agent-arg prompts work again while codex plugin stays denied. Robustness: a checked-in fixture (tests/fixtures/codex-subcommands.txt) of the full pinned subcommand list + a verify-invariants completeness check fail CI if a future pin bump adds an UNCLASSIFIED subcommand (which would otherwise leak as a prompt) — proven both directions. (P1) added a -c?* case that strips the glued short-flag prefix + optional = and runs the config blocker, closing -cfeatures.remote_plugin=true / -cfeatures={...} / -c=features.plugins=true while -cmodel=x still passes. New colima regressions: bare-prompt session permitted (SMOKE_EXIT=0), attached-c blocked. 27/27 unit, full go test, manifest verify on committed tree, real colima smoke+session green.

omkhar:
@codex review

omkhar:
Addressed in 643f960a: (P1) the completeness fixture is now derived from the pinned clap enum Subcommand SOURCE (openai/codex rust-v0.142.4 codex-rs/cli/src/main.rs), not codex --help — so the hidden/aliased dispatch tokens --help omits are covered: added cloud-tasks (alias of cloud), responses-api-proxy and stdio-to-uds (both #[clap(hide=true)]) to the fixture and the DENY case (exact-token, every UI). update dropped from the fixture (not a 0.142.4 variant; kept in the DENY case as harmless forward-compat since the check asserts fixture ⊆ classified). Completeness check proven both directions. (P2) codex_args_include_profile now stops at a bare -- so codex -- --profile breakglass keeps the managed --profile injected (verified via a bind-echo colima assertion of the injected argv); sibling argv scanners audited — codex_first_subcommand already breaks at --, no other helper affected. 27-case unit harness, full go test, verify-invariants completeness+manifest, real colima smoke (SMOKE_EXIT=0) + session (SCENARIO_EXIT=0) all green.

omkhar:
@codex review

omkhar:
Addressed in ff169e7a: you were right — the pinned 0.142.4 enum Subcommand DOES include a paren-less unit variant Update (between Completion and Doctor); my source extraction missed it and wrongly dropped it from the fixture. update is now in tests/fixtures/codex-subcommands.txt (it was already in the provider-policy DENY case), so the completeness check now genuinely enforces that the self-update surface stays classified; removed the incorrect no-update header note. pre-merge pr-parity green on the committed tree.

omkhar:
@codex review

omkhar:
Addressed in dbcb62fd: the fixture now carries a machine-readable # codex-version: 0.142.4 stamp, and verify-invariants asserts it EQUALS Dockerfile ARG CODEX_VERSION *before* the subset check — so a CODEX_VERSION bump that skips fixture regeneration fails closed here (with instructions to regenerate from the clap enum Subcommand source and bump the stamp) instead of the subset check passing vacuously on a stale list. Proven both directions: stamp==0.142.4 → verify-invariants exit 0; temporarily setting Dockerfile to the queued 0.144.1 → exit 1 at exactly this check. This makes the deny-by-default fence bump-safe. go test, verify-invariants end-to-end, real colima smoke (SMOKE_EXIT=0) + session (SCENARIO_EXIT=0) all green.

omkhar:
@codex review

omkhar:
@codex review

omkhar:
Addressed in e9c25e7f: the GUI app-server allowance now scans the app-server args and denies the control surface — the hidden --remote-control flag and every AppServerSubcommand/AppServerDaemonSubcommand token (daemon, proxy, generate-ts, generate-json-schema, generate-internal-json-schema, bootstrap, start, restart, enable-remote-control, disable-remote-control, stop, version, pid-update-loop) enumerated from the pinned rust-v0.142.4 source. The managed launch stays permitted because it is bare codex app-server (entrypoint.sh:220) and the sandbox/approval flags are prepended by provider-wrapper AFTER this check, so the post-app-server scan is empty for the managed path. AGENT_UI=gui codex app-server daemon enable-remote-control now denied; managed gui launch still works. Unit harness, go test, manifest+completeness+version-stamp checks, real colima smoke (SMOKE_EXIT=0) + session (SCENARIO_EXIT=0) green.

omkhar:
@codex review

omkhar:
@codex review

omkhar:
@codex review

omkhar:
@codex review

omkhar:
Addressed in 5488b878: extracted the top-level guarded-key set into codex_config_key_is_guarded(), and codex_config_override_is_blocked now, after normalizing, ALSO strips a profiles.<name>. prefix and re-tests the remainder against the same helper — so any key dangerous at top level is equally blocked profile-scoped (profiles.strict.features.remote_plugin, profiles.x.plugins.*, profiles.x.marketplaces.*, profiles.x.mcp_servers={...}, etc.). This is DRY and future-proof (a profile-v2 layer can set any key); the old hand-maintained profiles.*.sandbox_mode/approval_policy/… patterns are now covered by the general mechanism (regression-tested). Quoted profile names normalize correctly; a dot inside a quoted profile name fails closed. 23/23 harness, go test, manifest+completeness+version-stamp, real colima smoke (SMOKE_EXIT=0) + session (SCENARIO_EXIT=0) green.

omkhar:
@codex review

omkhar:
Simplify pass in a818e54a: condensed the accumulated per-round justification comments + verbose test-scaffolding headers to land under the 1200-line shape cap. Behavior is byte-identical to the reviewed 5488b878 — every deny/allow case, token, fixture entry, test assertion, and message string is unchanged (non-comment code diffs empty vs the prior head); only the control-plane manifest sha256 for provider-policy.sh rehashed. pre-merge pr-parity EXIT=0, real colima smoke+session green.

omkhar:
@codex review

omkhar:
Addressed in 80413cf0: replaced the app-server denylist with deny-by-default — once app-server is the permitted GUI first-subcommand, ANY following token dies (placed before the -- break so post-app-server prompt-position tokens die too; verified codex app-server has no [PROMPT] positional — app-server -- foo errors as an unrecognized subcommand). The managed launch reaches the check as bare app-server (sandbox flags prepended after), so it passes while --listen ws://IP:PORT, --stdio, --strict-config, daemon/proxy/generate-*, -c, and any future flag all fail closed without a hand-maintained list. Tests updated (bare managed launch permitted; --listen/--stdio/--strict-config/--help/daemon-enable-remote-control denied). check-pr-shape passes (1086), go test, real colima smoke (SMOKE_EXIT=0) + session (SCENARIO_EXIT=0) green.

omkhar:
@codex review

omkhar:
Addressed in a4cd40d0 (defense-in-depth): the features-action check now takes precedence over the -- break — while expect_features_action is armed, a -- is skipped without disarming so the following enable/disable is still denied; the normal prompt -- path is unchanged. Empirical note for the record: on the pinned Codex family (verified 0.142.4 and 0.144.1), codex features -- enable plugins is actually REJECTED by clap today ('subcommand enable exists; to use it, remove the -- before it') and does NOT mutate CODEX_HOME/config.toml — because features requires its <COMMAND> slot, unlike app-server's optional [COMMAND]. So this specific bypass is not live on the current pin; the reorder is applied as hardening so a future pin that makes features's command optional cannot reopen it. Sibling audit: no other --skips-a-pending-check path (expect_value consumes -- as a flag value and scans it; app-server dies on any arg). 15/15 unit, go test, manifest+completeness+version-stamp, real colima smoke (SMOKE_EXIT=0)+session (SCENARIO_EXIT=0), pre-merge EXIT=0, check-pr-shape 1157<1200.

omkhar:
@codex review

omkhar:
Addressed in dad1ab27: updated adapters/codex/README.md unsafe-argument policy to match the implemented reject_unsafe_codex_args — deny-by-default over the pinned Codex subcommand namespace (allowlist of read-only/session subcommands; plugin/remote-control/exec-server/mcp/mcp-server/cloud/cloud-tasks/responses-api-proxy/stdio-to-uds/update/sandbox/features enable|disable and any unclassified subcommand denied; bare-only GUI app-server; unknown first token = prompt), the version-stamped fixture pin, and the expanded flag/config policy (--yolo + --dangerously-bypass-* family; guarded -c namespaces incl. glued/quoted/escaped/inline-table/profile-scoped forms). check-doc-links + codespell clean.

omkhar:
@codex review

omkhar:
Addressed in 7cc16225 (two changes): (1) --remote-auth-token-env is now DENIED (bare and =value forms, alongside --remote) instead of value-consumed — it names the bearer-token env for the remote app-server websocket, a remote-auth surface; README updated to match, and unit+smoke assertions added. (2) Second simplify pass: condensed the accumulated per-round comments and collapsed repeated container-smoke assertion boilerplate into assert_codex_entrypoint_denied/assert_gui_appserver_denied helpers to land back under the 1200-line pre-merge cap (standalone shape 1078). Every distinct deny/permit test case is preserved (verified: --yolo, --dangerously-bypass-*, features={...}/hooks/marketplaces/profiles.*.features config bypasses, --listen, features -- enable, model gpt-5 plugin, etc. all still asserted), and every deny/allow case, token, fixture entry, and message string is byte-identical. go test, manifest+completeness+version-stamp, real colima smoke+session green.

omkhar:
@codex review

omkhar:
Addressed in 70a025dd: attached/glued short value-flags now apply the same guard as the space-separated forms — -a?* dies, -C?* dies (--cd), -p?* runs the profile check, -s?* runs the sandbox check (via two DRY helpers reused by the space-separated, --flag=value, and glued branches so all forms emit byte-identical messages); case-sensitive placement so -C?* never shadows the config -c?*. Long --profile=/--sandbox=/--cd= forms audited (already covered). Also fixes the broken permit test that failed CI container-smoke on the prior commit (-pstrict/--profile=strict were asserted via clean-exit, but codex itself exits nonzero on a duplicate managed --profile — now asserted via absence of the Workcell block message, matching real container behavior). 17/17 unit, go test, manifest+completeness+version-stamp, real colima smoke (SMOKE_EXIT=0)+session (SCENARIO_EXIT=0), shape 1156.

omkhar:
@codex review

omkhar:
Addressed in fe002fba: the guarded value helpers (codex_reject_unsafe_profile_value / codex_reject_unsafe_sandbox_value) now strip one leading = before comparing, so the clap short-with-equals form -s=danger-full-access / -p=breakglass is checked exactly like -sdanger-full-access and -s danger-full-access (the glued -p?*/-s?* cases previously passed =VALUE straight into the exact comparison, which missed it and forwarded -s=danger-full-access). No-op for the space-separated and --flag= callers (their values never lead with =). Unit-verified: -s=danger-full-access and -p=breakglass now die; -s=read-only and -p=strict permitted. Smoke assertions added for both short-equals forms.

omkhar:
@codex review

omkhar:
@codex review

omkhar:
Addressed in 6c7ff772: debug is denied wholesale — it was allowlisted as read-only, but its second-level subcommands are not (debug app-server reaches the app-server test client; debug clear-memories mutates local memory) and the gate only checks the first token. The managed path never uses debug (grep-verified across scripts/runtime/tests/internal), so it moves from ALLOW to the DENY set; it remains in the version-stamped fixture (now classified deny, completeness check still passes). README updated. Unit-verified: debug and debug app-server send-message-v2 denied, exec still permitted. Smoke assertions added; go test + lint green.

omkhar:
@codex review

omkhar:
@codex review

omkhar:
@codex review

omkhar:
@codex review

PR #397, opened 2026-07-05
^d D4: complete launcher-contract.md with required-tools/env/exit-codes/test-flags reference

omkhar:
**D4 (modularize the launcher) — completes the contract doc.** Follows the three module extractions (#394 host-detect, #395 host-exec, #396 go-hostutil). D4's acceptance criteria require `docs/launcher-contract.md` to document **required tools, environment expectations, exit codes, and test override flags** — this adds a "Launcher reference" block with those four sections.

## What
A doc-only section (110 lines) after the intro:
- **Required host tools** — the two resolvers (`resolve_fixed_host_tool` fixed-path / `resolve_host_tool{,_optional}` candidates+trusted-PATH), and each tool (`go`, `colima`, `docker`, `git`, `curl`) with its purpose; abort-on-missing contract.
- **Environment expectations** — the pinned `TRUSTED_HOST_PATH`, the full `scrub_host_process_env` unset list, and derived vars (`REAL_HOME`, `WORKCELL_GO_CACHE_ROOT`, `COLIMA_STATE_ROOT`/`HOME`, `GOPATH`/`GOMODCACHE`/`GOCACHE`) + the go-hostutil ordering invariant.
- **Exit codes** — the four actually used (`0`/`1`/`2`/`88`), grouped by meaning.
- **Test override flags** — each `WORKCELL_TEST_*`/`WORKCELL_VERIFY_*` flag, its effect, and its gating (harness-only; e.g. reject-on-presence guards for the credential-file vars, and the env-vs-internal split on the fault-injection flag).

## Accuracy
Every claim is grep-derived from `scripts/workcell` + the launcher modules (not assumed); I independently re-verified the exit-code set (0/1/2/88), the eight test/verify flags, the tool resolvers, and the reject-guard line numbers against source before pushing.

## Validation
markdownlint clean. Doc-only — no code touched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #362, opened 2026-07-03
Add GitHub Copilot Tier 1 adapter

omkhar:
## Summary
- Add the GitHub Copilot Tier 1 adapter, launcher/runtime integration, auth handoff, and provider policy controls.
- Document Copilot support, requirements, operator contract, validation scenarios, release/pin verifier paths, and support matrices.
- Add Copilot e2e/dry-run coverage and upstream release verification.

## Validation
- Live certification passed earlier on this Mac: `WORKCELL_COLIMA_START_TIMEOUT_SECONDS=600 ./scripts/provider-e2e.sh --agent copilot --workspace "$PWD" --require-injection`.
- Local gates passed before signing: `./scripts/verify-upstream-copilot-release.sh`, `./scripts/container-smoke.sh`, `./scripts/verify-invariants.sh`, `./scripts/validate-repo.sh`.
- Fresh rebased PR parity passed: `./scripts/pre-merge.sh --profile pr-parity --label approved-large-certified-adapter`.
- Adversarial multi-agent review completed before publication; actionable findings were fixed.

PR #359, opened 2026-07-02
Harden Copilot release verification modes

omkhar:
## Summary

- Add explicit Copilot release verifier help modes for native, Docker, checksum-only, and auto behavior.
- Keep local validation, pin hygiene, and provider bump paths on checksum-only verification so they do not require a prepared runtime image.
- Preserve runtime-image verification for callers that explicitly select native/Docker help modes.

## Validation

- PASS: WORKCELL_GITHUB_API_TOKEN_FILE=<redacted> WORKCELL_COPILOT_RELEASE_HELP_MODE=checksum ./scripts/verify-upstream-copilot-release.sh
- PASS: bash -n scripts/verify-upstream-copilot-release.sh scripts/ci/job-pin-hygiene.sh scripts/ci/job-validate.sh scripts/update-provider-pins.sh
- PASS: ./scripts/check-pr-shape.sh --base-ref origin/main --head-ref HEAD --max-files 25 --max-lines 1200 --max-areas 8 --max-binaries 0
- PASS: WORKCELL_COLIMA_START_TIMEOUT_SECONDS=600 ./scripts/pre-merge.sh --profile pr-parity

PR #493, opened 2026-07-12
Close out ambiguous historical Codex findings: deny Copilot logout + document buildx override

omkhar:
## Summary

Closes out the six ambiguous historical Codex findings that were left unreacted on PRs #336, #359, #362, and #397. I verified each against the current tree (not the PR-era code) and reacted 👍 on all six. Verdicts:

| Finding | PR | Verdict |
| --- | --- | --- |
| codex `--dangerously-bypass-hook-trust` forwarded | #336 | **Already fixed** — current guard denies `--dangerously-bypass-*` (verified against the pinned codex binary). |
| app-server missing `approval_policy` | #336 | **Handled by design** — app-server gets its approval policy from the injected `--ask-for-approval` flag (documented in the wrapper). |
| copilot help verification skipped in checksum-only mode | #359 | **Addressed** — the default mode is now `auto` (does native/docker help verification), and release callers set `native`/`docker` explicitly; no bare caller selects `checksum`. |
| copilot business/enterprise egress hosts absent | #362 | **Deliberate scoping** — the default strict allowlist intentionally covers only `api.githubcopilot.com` + `api.individual.githubcopilot.com` + github hosts; adding wildcard `*.business.githubcopilot.com` would broaden default egress. Not changed. |
| **copilot `logout` not denied** | #362 | **Live gap — fixed here.** |
| `WORKCELL_TRUSTED_BUILDX_BIN` override undocumented | #397 | **Doc-only — added here.** |

## Changes

1. **`^F` copilot `logout` deny** — `reject_unsafe_copilot_args` blocked `login`/`init`/`mcp`/`plugin`/`skill`/`update` but omitted `logout`, so a managed Copilot session could clear persisted auth state — asymmetric with the denied `login`. Added `logout` to the lifecycle deny list, with a container-smoke denial assertion, and regenerated the control-plane manifest.

2. **`^d` buildx override doc** — documents in `docs/launcher-contract.md` that a pre-set executable `WORKCELL_TRUSTED_BUILDX_BIN` short-circuits the trusted-plugin search but is cleared by the launcher's `env -i` shebang on any normal invocation, surviving only for shebang-bypass `/bin/bash -p scripts/workcell` runs that are already outside the trust boundary.

## Validation

- `go test ./...`, `scripts/container-smoke.sh` (colima rebuild, new `logout` assertion), `scripts/verify-invariants.sh` — all green.
- markdownlint + codespell + check-doc-links clean.
- `scripts/pre-merge.sh --profile pr-parity` — exit 0.

### DIFF
--- a/adapters/copilot/README.md
+++ b/adapters/copilot/README.md
@@ -49,8 +49,8 @@ directories. In-container reserved session targets: `~/.copilot`,
   §3).
 - Unsafe-argument policy (`reject_unsafe_copilot_args` in
   `runtime/container/provider-policy.sh`): the wrapper blocks the
-  `init`/`login`/`mcp`/`plugin`/`skill`/`update` lifecycle and control-plane
-  subcommands and a broad set of trust-widening flags (MCP toolset/config,
+  `init`/`login`/`logout`/`mcp`/`plugin`/`skill`/`update` lifecycle and
+  control-plane subcommands and a broad set of trust-widening flags (MCP toolset/config,
   `--allow-all*`, `--allow-tool`/`--allow-url`, `--add-dir`, `--dynamic-retrieval`,
   `--no-custom-instructions`, `--remote`/`--share*`/`--worktree`, `--yolo`,
   bundled short options, and more). These are rejected in **every** mode
--- a/docs/launcher-contract.md
+++ b/docs/launcher-contract.md
@@ -82,7 +82,13 @@ prerequisites sit outside this table entirely. Remote-preview backends probe ext
 `session-manager-plugin` for the `aws-ec2-ssm` backend, `gcloud` for `gcp-vm`.
 Image builds require a system `docker-buildx` plugin binary
 (`ensure_workcell_trusted_buildx`, `scripts/lib/trusted-docker-client.sh`), which
-aborts with `Missing trusted docker-buildx binary` when none is found. And the
+aborts with `Missing trusted docker-buildx binary` when none is found. A
+pre-set, executable `WORKCELL_TRUSTED_BUILDX_BIN` short-circuits that search and
+is honored as-is; the launcher's `env -i` shebang (line 1) clears it on any
+normal invocation, so this override survives only for shebang-bypassed
+invocations (`/bin/bash -p scripts/workcell`) that inherit an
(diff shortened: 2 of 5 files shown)


omkhar:
@codex review

omkhar:
Addressed in c9ba60cb365d73d84ab53c9f8a928b46b400b3b2: synced adapters/copilot/README.md to list `logout` alongside the other blocked lifecycle subcommands; the README is an adapter-baseline control-plane input so the manifest was regenerated in lockstep. Verified the P3 against source (the README line did omit logout).

omkhar:
@codex review

omkhar:
Re the P2 `copilot -p "/logout"` slash-command finding — reacting 👍 but scoping it out of this PR with reasoning rather than adding a narrow filter:

1. **What this PR fixes** is the confirmed `logout` *subcommand* gap (`copilot logout`), asymmetric with the already-denied `login`; verified against the pinned 1.0.70 binary + container-smoke assertion.

2. **The `-p "/logout"` slash vector is unverifiable here**: the managed entrypoint requires an auth token-handoff file, and the real binary returns `No authentication information found` before any prompt processing, so I can't empirically confirm whether non-interactive `-p` even executes a leading-slash prompt as a control command (the bundled Node binary's app logic isn't strings-introspectable either). Per this repo's empirical-evidence rule I won't claim it fixed or refuted without an authenticated repro.

3. **If real, a `/logout`-string filter is incomplete**: the same path would expose `/login`, `/reset`, `/model`, etc., and prompt content is free-form data (whitespace / multiline / `/logout arg` evade a substring match). The proportionate fix is a copilot flag disabling slash processing in `-p` mode (none exists in 1.0.70 `--help`) or a deliberate design decision on slash reachability — a distinct, broader change.

4. **Blast radius is self-limited**: `-p` is set at launch (the agent doesn't re-invoke copilot); clearing a session's auth view is session-local (the host-staged handoff file persists; relaunch re-stages). No exfiltration, escape, or persistence.

Tracking the slash-via-prompt question as a follow-up needing an authenticated repro; not blocking the confirmed subcommand fix.

omkhar:
@codex review

PR #521, opened 2026-07-23
Add descriptor-safe session state primitives

omkhar:
## Summary

- add descriptor-relative, symlink-resistant directory walking for owner-controlled session state
- add bounded owner-only regular-file reads with identity and metadata revalidation
- add deterministic syscall seams and adversarial race, ownership, mode, link-count, close, and platform tests

This is a dormant package-private foundation for the sequenced C3 shared-profile work. It does not activate parallel sessions, change the CLI or public contract, or make a new support claim.

## Risk and assurance

- the implementation is Unix-only and unreachable from shipped runtime paths in this unit
- directory traversal is rooted at a trusted descriptor and fails closed on noncanonical paths, symlinks, unsafe ancestors, and same-read mutations
- same-UID mutation outside the enclosing state lock is explicitly outside these helpers' guarantee

## Validation

- independent adversarial review reached clean consensus on the exact rebased tree
- focused/full Go tests, race detector, vet, staticcheck, and 20 shuffled runs passed
- Darwin and Linux `amd64`/`arm64` compilation passed
- hardened read-only Linux execution passed as root and UID 65532
- full `pr-parity` passed on the exact staged tree, including live Colima invariants and release-payload checks

No live C3 support certification is required for this dormant, unreachable foundation; later behavior-changing units remain fail closed until their exact-tree live gates pass.

### DIFF
--- a/internal/host/sessions/owner_file.go
+++ b/internal/host/sessions/owner_file.go
@@ -1,0 +1,152 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+package sessions
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"math"
+	"path/filepath"
+	"strings"
+
+	"golang.org/x/sys/unix"
+)
+
+type ownerFile struct {
+	data []byte
+	stat unix.Stat_t
+}
+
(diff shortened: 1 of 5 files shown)


omkhar:
@codex review

PR #524, opened 2026-07-23
Fix canonical hosted-runner Go aliases

omkhar:
## Summary

- scrub only passive hosted-runner aliases matching `GOROOT_<major>_<minor>_{X64,ARM64}` before canonical descendants run
- continue rejecting `GOROOT`, malformed or unknown aliases, and every active ambient Go selector
- fail closed if an alias cannot be removed, without exposing its value
- bind the narrowed behavior in regression, mutation, Bash-version, and documentation-reachability tests

## Why

Merged-main Hosted controls run [29996637812](https://github.com/omkhar/workcell/actions/runs/29996637812) failed because the GitHub runner supplies `GOROOT_1_24_X64`; that versioned tool-cache alias is passive metadata rather than a Go tool input. The original blanket `GO*` gate therefore rejected a normal hosted environment.

## Validation

- focused normal and race tests for the canonical environment
- Bash 3.2 and Bash 5.3 alias, near-match, idempotence, and readonly-failure probes
- three killed mutations: classifier removal, scrub removal, and ignored `unset` failure
- ShellCheck, shfmt, diff, and documentation reachability checks
- live repository hosted-controls audit with the exact failing alias injected
- full exact-tree `pr-parity`
- independent exact-tree review clean after both review findings were fixed

An exact-head manual Hosted controls dispatch was attempted in
[run 30001465746](https://github.com/omkhar/workcell/actions/runs/30001465746),
but the `hosted-controls-audit` environment correctly rejected the PR branch
before checkout: its secret-bearing policy permits only `main` and `v*` tags.
That environment will not be weakened for a PR. The exact tree instead has the
successful live read-only audit above, and the merged-`main` Hosted controls run
must pass before this repair is considered complete.

### DIFF
--- a/docs/validation-scenarios.md
+++ b/docs/validation-scenarios.md
@@ -81,7 +81,10 @@ explicitly source `scripts/lib/canonical-build-env.sh`.
 
 The gate rejects every nonempty ambient shell-identifier `GO*` and `CGO*`
 variable except exact `GOENV=off`, exact `GOWORK=off`, empty `GOFLAGS`, and
-caller-selected `GOPATH`, `GOCACHE`, and `GOMODCACHE` storage paths. It also
+caller-selected `GOPATH`, `GOCACHE`, and `GOMODCACHE` storage paths. Passive
+hosted-runner tool-cache aliases matching exact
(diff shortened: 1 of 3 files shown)


omkhar:
@codex review

omkhar:
@codex review

PR #639, opened 2026-09-06
^R Deduplicate smoke wrappers and delete redundant host probes (ponytail audit batch 2)

omkhar:
Batch 2 of the ponytail-audit remediation: the four verified-safe shell cuts.

- container-smoke.sh: the nine run_* docker-run wrappers (30 identical lines each) collapse into one smoke_run_container core plus thin named wrappers. All nine public names, argument contracts, and flag/env/mount semantics are unchanged, including the deliberate absence of the workspace-mutable-exec skip knob in the injection-bundle variants.
- verify-invariants.sh: the nine standalone hostile-BASH_ENV/BASH_FUNC/PATH probe blocks are deleted; all three probed scripts are HOST_GATE_SCRIPTS members, so the gate loops already run byte-identical probes plus a hostile-bash probe the standalone blocks lacked. The script's own self-probes stay.
- bootstrap-dev.sh: the 68-line flag parser no caller uses is gone; the script now runs dev-quick-check.sh to verify the setup.
- The four identical select_docker_context() wrappers are inlined at their call sites.

The re-scope that produced this short list is deliberate: the audit's larger shell-dedup items (require_tool, env -i preambles, resolve_go_bin, runtime wrapper env scrubs) were all verified to be enforced hardening architecture and were rejected, not deferred.

Evidence: go test ./... 47 packages ok (including the suites that pin these scripts' literal text); shellcheck -x and shfmt clean; pre-merge --profile pr-parity pass (runs the real container smoke through the consolidated wrappers).

https://claude.ai/code/session_01DBuWTCc8rBa6gbNzpqpTYJ

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ Use the bootstrap helper:
 ```
 
 That script installs the common local toolchain, configures `.githooks` as the
-repo hook path, and leaves you ready to run the local gates.
+repo hook path, and runs `./scripts/dev-quick-check.sh` to verify the setup.
 
 ## Prerequisites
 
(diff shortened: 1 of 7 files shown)


omkhar:
@codex review

PR #649, opened 2026-09-06
^R Port the Codex TOML invariants to Go and split the live Colima lane

omkhar:
Three related reductions to `scripts/verify-invariants.sh`: the hand-rolled TOML parser goes away, the static Go delegations stop paying per-check process spawns, and the Darwin live-Colima lane moves to its own file.

## Summary

- Port the Codex TOML invariants to Go. `internal/metadatautil.ValidateCodexManagedConfig` / `ValidateCodexProfileLayer` / `ValidateCodexAdapterLockstep` replace the ~600-line awk `toml_section_assignments` / `toml_section_names` parser and its `require_toml_*` helpers, decoding through the already-vendored `BurntSushi/toml`. The new `workcell-codex-toml-invariants` subcommand runs them in the original script order (both managed baselines, routing parity, the four profile-v2 layers, requirements/wrapper lockstep), so first-failure ordering, exit codes, and stderr text are unchanged. Comparison is on decoded values rather than raw assignment text: it accepts benign reformatting (`enabled=true` vs `enabled = true`) and never a different value; undecodable configs still fail closed.
- Carry the former in-script mutant fixtures over as Go unit tests, including the positive control that a quoted single-segment section name with literal dots (`["profiles.strict.x"]`) decodes to a distinct top-level key and must not trip the `[profiles...]` guard.
- Batch the static Go delegations. `workcell-check-batch ROOT_DIR CHECK[=ARG] ...` runs a contiguous run of single-argument checks in argv order inside one process, stopping at the first failure so its exit code and stderr are byte-identical to the individual subcommand. Batchable names are resolved from the existing subcommand table, so there is no second registry to keep in sync. 48 `workcell-citools` spawns become 19.
- Split the Darwin live-Colima lane into `scripts/verify-invariants-live.sh`, sourced (not executed) so the existing traps and `ROOT_DIR` scope are preserved. `verify-invariants.sh` drops from 8697 to 7193 lines.
- Collapse the new Go helpers onto stdlib: `reflect.DeepEqual` replaces a hand-rolled value comparator (which also removes a latent panic on nested arrays), `slices.Sorted(maps.Keys(...))` replaces the manual key-collect-and-sort, and the two citools test process helpers become one.

## Testing

- `go build ./...` — clean.
- `go vet ./...` — clean.
- `go test ./... -count=1` — exit 0, all packages ok.
- `go test ./internal/metadatautil/... ./cmd/workcell-citools/... -count=1` — ok (metadatautil 7.6s, workcell-citools 0.8s), covering the mutant fixtures, the profile-layer rejections, the lockstep fixtures, and the batch ordering / usage-error cases.
- `go run ./cmd/workcell-citools workcell-codex-toml-invariants <repo>` — exit 0 against the shipped adapter configs.
- `shellcheck -x scripts/verify-invariants-live.sh` — clean; `bash -n` clean on both `verify-invariants.sh` and `verify-invariants-live.sh`.
- Confirmed `tests/scenarios/shared/test-session-commands.sh` still extracts both `detached_session_monitor_*` functions from their new home.


omkhar:
Superseded by a unit series, to satisfy the PR shape gate. This branch is +1710/-1620, well over the 1200-changed-line limit, so the work ships as smaller units that each build and test green on their own and each leave `scripts/verify-invariants.sh` fully functional.

The first unit is #667 (the Go port package plus the `workcell-codex-toml-invariants` citools subcommand, additive only, 3 files / +693). The remaining units follow as their predecessors merge:

- **Unit 2** — swap the `verify-invariants.sh` Codex TOML region for a call to the new subcommand and delete the awk TOML parser and the `require_toml_*` helpers (~671 changed lines, 1 file).
- **Unit 3** — add the `workcell-check-batch` subcommand and batch the static Go delegations in `verify-invariants.sh` (~470 changed lines, 4 files).
- **Units 4a/4b** — extract the Darwin live-Colima lane into `scripts/verify-invariants-live.sh`. A pure move counts both the addition and the deletion, so the single-shot move is ~1506 changed lines and has to land in two steps, each moving roughly half the lane.

Keeping `codex/invariants-toml-port-v2` around until the series lands, as the reference for the remaining extractions.

PR #650, opened 2026-09-06
Verify release outputs after signing

omkhar:
## Summary

- Add `scripts/verify-release-outputs.sh` and an independent read-only `verify-release-outputs` job that re-verifies every release output after the signing job produces it: the Sigstore bundle for each of the nine release-data assets, the runtime image signature, the release and `sha-<commit>` image tags against the signed digest, and the eight provenance plus two SBOM attestations.
- Require the publisher to repeat the same checks before it uploads the assets, so a change between sealing and publication fails closed.
- Move the release trigger from the pushed tag to `repository_dispatch`, so the workflow always loads from `main`, and bind every downstream job to the release commit that the tag-policy gate verified.
- Extend the tag-policy gate to check the workflow ref, the workflow SHA, the tag target, and the required `main` workflow set. This subsumes the previous ancestry and check-run gates, which are removed.
- Bind the publisher to the annotated tag object with `--expected-tag-object`, read the release commit only from the canonical tag headers instead of peeling, and pin the Go toolchain to the exact `go.mod` version at a trusted absolute path.
- Make attestation unconditional. The mutable `WORKCELL_RELEASE_NO_ATTEST` and `WORKCELL_ENABLE_PRIVATE_GITHUB_ATTESTATIONS` repository-variable opt-out is gone, replaced by a fixed public-visibility guard that the validator matches exactly.
- Update the workflow validators, lane policy, and release documentation for the dispatch trigger and the `main` workflow identity.

The second commit is an over-engineering pass over the first: it removes eight job-level `RELEASE_TAG` and `RELEASE_COMMIT` bindings that no step reads, folds the unconditional `verify_args+=(--attestations)` append into the array literal, deletes `namedWorkflowStep` and its test after the YAML-based attestation validator replaced it, collapses two duplicated image-digest rejection tests into one table, and drops a substring permission scan that the parsed permission-map assertion already covers exhaustively. Net 71 lines removed with no coverage lost.

## Testing

- `go build ./...` passed.
- `gofmt -l internal/` reported no files.
- `go vet ./internal/metadatautil ./internal/host/release` passed.
- `go test ./internal/metadatautil -count=1` passed in 14.1s.
- `go test ./internal/host/release -count=1` passed in 2.9s.
- `go test ./internal/testkit -run VerifyReleaseOutputs -count=1 -v` passed, 15 tests including the Cosign, attestation, checksum-inventory, symlink, moved-tag, and malformed-digest negative controls.
- `./scripts/check-pinned-inputs.sh` exited 0.
- `./scripts/verify-workflow-lanes.sh` exited 0.
- `./scripts/check-workflows.sh` exited 0; zizmor v1.23.1 reported no findings.
- `./scripts/check-dead-code.sh` passed. Its one report, `internal/injectionpolicy/reader_acl.go:10 isPolicyACLName`, predates this branch and is untouched here.

The release workflow itself is hosted-only and remains subject to GitHub validation on the next dispatched release.


omkhar:
Superseded: this PR exceeds the repository PR shape gate (16 files, +1788/-366 across the release workflow, its validators, policy, docs, and two new test suites), so it is being split into reviewable units that each build and test green alone.

The first unit is #669, which carries `scripts/verify-release-outputs.sh`, the independent `verify-release-outputs` job, the publisher re-verification step, the publication-gate validator changes, and the workflow-lane registration. The remaining units carry the verifier's behavioural test suite, the `repository_dispatch` trigger and release-commit binding, the unconditional attestation flow, and the documentation updates.

Closing this PR. The branch `security/release-output-verification` is kept as the extraction source until the series lands.

PR #651, opened 2026-09-06
Replace the shell apt broker with a Go broker

omkhar:
Replaces `runtime/container/apt-broker.sh` and the sudo-wrapper's inline request/result directory protocol with a root-owned Go broker over a Unix socket.

## Summary

- `cmd/workcell-apt-broker-server` listens on `/run/workcell/apt-broker/socket` (root-owned dir `0755`, socket `0666`) and runs the fixed package helper; `runtime/container/bin/sudo-wrapper.sh` collapses from 222 lines to a five-line exec into `cmd/workcell-apt-broker-client`.
- Admission is the kernel-attested `SO_PEERCRED` uid matched against the mapped runtime user, so the wire protocol carries only an argument vector and an allowlisted environment — the client cannot choose what runs. Request size, response size, concurrency, and helper duration are all bounded.
- The per-session `sudoers.d` NOPASSWD entry is dropped with the shell broker that needed it. Startup moves from `entrypoint.sh` into `runtime-user.sh`, which passes the mapped runtime uid to the server before handing control to that user, and refuses to reuse or remove an existing socket path.
- A new pinned Dockerfile stage builds both binaries from the pinned Go toolchain; the runtime build context, outbound allowlist, and upstream pin refresh gain the Go toolchain inputs. Codex artifacts install root-owned with explicit modes.
- Second commit is an over-engineering pass over the new sources: `writeFull` (the `io.Writer` contract already forbids a silent short write) and its two fakes, a per-argument length check the body-size bound already covers, two `bytes.Reader` aliases, a duplicated socket predicate, a restated directory validator, a redundant `ErrorWriter` default, and a one-method anonymous interface. The sudo shim now names `aptbroker.DefaultHelperPath` instead of repeating its literal. No boundary behavior changes.
- `docs/invariants.md` gains invariant 4b for the broker; `docs/outbound-endpoints.md` and `policy/hardening-profile.toml` gain the Go toolchain endpoints.

## Testing

- `go build ./...` — clean
- `GOOS=linux go build ./...` — clean
- `go vet ./internal/aptbroker/ ./cmd/...` — clean
- `GOOS=linux go vet ./internal/aptbroker/ ./cmd/...` — clean (covers the linux-only helper, peer, and startup files and their tests)
- `gofmt -l internal cmd` — no output
- `go test ./... -count=1` — all packages ok except one rerun-clean flake in `internal/host/c3certify` (`TestKeepaliveHandlesSignals/terminated`, signal timing, unrelated to this change); `go test ./internal/host/c3certify/ -run TestKeepaliveHandlesSignals -count=2` passes
- `go test ./internal/aptbroker/ ./cmd/workcell-apt-broker-client/ ./cmd/workcell-apt-broker-server/ -count=1` — ok, run again after each cleanup commit
- Container smoke covers the broker end to end: `run_slow_apt_broker_smoke` starts the server as root under `env -i`, asserts socket ownership and mode, then drives an 11-second helper through `sudo -n` as uid 12345 and requires the request to wait rather than time out.


omkhar:
Closing in favour of a smaller series. This branch is +4807/-574 across 61 files, which is well over the repo's PR shape limit, so it is being landed as six reviewable units that each build and test green on their own.

The container keeps a working apt broker at every point: `runtime/container/apt-broker.sh` and the existing sudo wrapper stay in place and unchanged until the final unit swaps them.

1. #668 — `internal/aptbroker` wire protocol and client (open)
2. `internal/aptbroker` privileged helper runner
3. `internal/aptbroker` socket server and `SO_PEERCRED` peer admission
4. `cmd/workcell-apt-broker-client` and `cmd/workcell-apt-broker-server`
5. Pinned Go toolchain build stage, build-context inputs, and outbound allowlist
6. Runtime swap: entrypoint and wrapper wiring, deletion of the shell broker, smoke and hardening tests

No content is dropped; the units together are this branch.

PR #652, opened 2026-09-06
!B Harden container launcher exec paths (Rust suite, clippy, and adapter self-check pass; user-visible execution boundary)

omkhar:
Hardens the two exec paths a child can use to arrive outside the guard, and adds the variadic exec adapter that the guard was missing.

## Summary

- `sanitize_env` restores the immutable Workcell guard preload after it clears caller-controlled loader state. Previously the sanitizer left `LD_PRELOAD` unset, so a child could arrive with a hostile `LD_PRELOAD` and escape the exec guard. The restored value is the fixed `GUARD_PRELOAD` constant, not anything the caller supplied.
- Adds `src/exec_variadic.c` and the `build.rs` step that compiles and archives it on native Linux builds. The adapter supplies real `execl`/`execlp`/`execle` implementations, which flatten their variadic argument lists and re-enter the guard through `workcell_execv`/`workcell_execvp`/`workcell_execve` rather than reaching libc directly.
- Links the guard cdylib with `-Wl,-Bsymbolic-functions` so an interposer cannot replace the guard's internal dispatch between the adapter and the Rust bridges.
- Splits each intercepted entry point into a `guarded_*` implementation plus thin exported symbols, so the adapter, the unversioned exports, and the syscall shim all funnel through one classifier.
- The signal-forwarding hardening from the same review already landed on `main`; this keeps `main`'s version, which is a superset.

## Notes for review

- `tests/exec_variadic_probe.c` is inert: nothing in `build.rs`, the Dockerfile, or `scripts/` compiles it yet. It is staged for the image build that will run it.
- The `WORKCELL_SYMVER` block in `exec_variadic.c` is likewise inert — `build.rs` is the only compiler of that file and always passes `-DWORKCELL_NO_SYMVER`. Flagged here rather than cut, since the versioned-export decision belongs with the image build step that would enable it.

## Testing

Run from `runtime/container/rust`:

- `cargo fmt --check` — clean.
- `cargo test` — `21 passed; 0 failed` (lib), `17 passed; 0 failed` (workcell-launcher), `0` doc-tests. Includes the updated `sanitize_env_clears_sensitive_loader_and_node_state`, which now asserts `LD_PRELOAD` equals `GUARD_PRELOAD` instead of being unset.
- `cargo clippy --all-targets` — zero warnings, zero errors (`undocumented_unsafe_blocks` is `deny` for this crate).
- `cc -std=c11 -fPIC -fvisibility=hidden -O2 -Wall -Wextra -Werror -DWORKCELL_NO_SYMVER -c src/exec_variadic.c` — compiles clean under `-Werror`; `nm -g` shows all 12 `workcell_export_*` symbols.
- Differential self-check on the argument collector: a stub harness binding `workcell_execv`/`workcell_execvp`/`workcell_execve` asserts argv flattening, the NULL sentinel, and the post-sentinel `execle` envp for arg0-only and multi-arg calls. Built against both the pre-dedup and post-dedup adapter; both report `collect_exec_args self-check OK`.

Host is macOS, so `build.rs` skips the C compile and the `#[cfg(target_os = "linux")]` guard paths are not exercised locally — those are covered by the Linux CI lane and the container smoke suite. The adapter was compiled and exercised manually as described above.

## Cleanup in this branch

- Merged `collect_exec_args` and `collect_exec_args_and_env`, which ran the same bounded scan and the same argv allocation, into one collector with an optional `envp` out-parameter. Behavior verified identical by the differential self-check above.
- Dropped a post-compile `fs::metadata` check on the adapter source in `build.rs`; the compiler had already read the file, so a missing source fails the compile above it.

### DIFF
--- a/runtime/container/rust/build.rs
+++ b/runtime/container/rust/build.rs
@@ -1,0 +1,62 @@
+use std::env;
+use std::fs;
+use std::path::PathBuf;
+use std::process::Command;
+
+fn main() {
+    println!("cargo:rerun-if-changed=src/exec_variadic.c");
+
+    if env::var("CARGO_CFG_TARGET_OS").as_deref() != Ok("linux")
+        || env::var("HOST").ok() != env::var("TARGET").ok()
+    {
+        return;
+    }
+
+    let out_dir = PathBuf::from(env::var_os("OUT_DIR").expect("Cargo sets OUT_DIR"));
+    let object = out_dir.join("exec_variadic.o");
+    let source = PathBuf::from("src/exec_variadic.c");
+    let compiler = env::var_os("CC").unwrap_or_else(|| "cc".into());
+    let status = Command::new(compiler)
+        .args([
+            "-std=c11",
+            "-fPIC",
+            "-fvisibility=hidden",
+            "-O2",
+            "-Wall",
+            "-Wextra",
+            "-Werror",
+            "-DWORKCELL_NO_SYMVER",
+            "-c",
+        ])
+        .arg(&source)
+        .arg("-o")
+        .arg(&object)
+        .status()
+        .expect("compile the Linux variadic exec adapter");
+    assert!(
+        status.success(),
+        "Linux variadic exec adapter compilation failed"
+    );
+
+    let archive = out_dir.join("libworkcell_exec_variadic.a");
+    let archiver = env::var_os("AR").unwrap_or_else(|| "ar".into());
+    let status = Command::new(archiver)
+        .args(["rcs"])
+        .arg(&archive)
+        .arg(&object)
+        .status()
+        .expect("archive the Linux variadic exec adapter");
+    assert!(
+        status.success(),
+        "Linux variadic exec adapter archiving failed"
+    );
+
+    // Link the archive into every native target. Unit tests must resolve the
+    // exported trampolines as well as the preload shared library.
+    println!("cargo:rustc-link-search=native={}", out_dir.display());
+    println!("cargo:rustc-link-lib=static=workcell_exec_variadic");
+    // Bind adapter calls to the Rust bridges in this shared object. This avoids
+    // a caller-controlled interposer replacing the guard's internal dispatch.
+    println!("cargo:rustc-link-arg-cdylib=-Wl,-Bsymbolic-functions");
+    let _ = fs::metadata(source).expect("variadic exec adapter source exists");
+}
--- a/runtime/container/rust/build.rs
+++ b/runtime/container/rust/build.rs
@@ -1,5 +1,4 @@
 use std::env;
-use std::fs;
 use std::path::PathBuf;
 use std::process::Command;
 
@@ -58,5 +57,4 @@ fn main() {
     // Bind adapter calls to the Rust bridges in this shared object. This avoids
     // a caller-controlled interposer replacing the guard's internal dispatch.
     println!("cargo:rustc-link-arg-cdylib=-Wl,-Bsymbolic-functions");
-    let _ = fs::metadata(source).expect("variadic exec adapter source exists");
 }
--- a/runtime/container/rust/src/bin/common/launcher_common.rs
+++ b/runtime/container/rust/src/bin/common/launcher_common.rs
@@ -15,6 +15,7 @@ unsafe extern "C" {
 }
 
 pub const BASH_PATH: &str = "/bin/bash";
+pub const GUARD_PRELOAD: &str = "/usr/local/lib/libworkcell_exec_guard.so";
 
 #[cfg(not(test))]
 static MANAGED_CHILD_PID: AtomicI32 = AtomicI32::new(0);
@@ -53,6 +54,9 @@ pub fn sanitize_env() {
         // SAFETY: called during single-threaded launcher startup before any thread or child is spawned; no concurrent env access.
         unsafe { env::remove_var(key) };
     }
+    // Restore only the immutable Workcell guard after removing caller-controlled loader state.
+    // SAFETY: called during single-threaded launcher startup before any thread or child is spawned.
+    unsafe { env::set_var("LD_PRELOAD", GUARD_PRELOAD) };
 }
 
 pub fn set_env_var(key: &str, value: &str) {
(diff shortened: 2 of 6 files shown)


omkhar:
Closing in favour of an incremental series. This branch is +2981/-160 with 2821 of those lines in a single file (`runtime/container/rust/src/lib.rs`, 2166 -> 4669 lines), which is past the PR shape gate and past what can be reviewed as one change.

A note for whoever picks up the follow-ups: the description above covers the guard-preload restore, the variadic adapter, `-Bsymbolic-functions` and the `guarded_*` split, but the diff also carries a substantially larger body of work that the description does not mention — roughly 2500 lines adding bounded exec-input parsing, a trusted-immutable path and descriptor provenance layer, untrusted-shebang classification, and a memfd snapshot-and-seal execution path for mutable targets, plus ~480 lines of tests for those. Each of those is its own hardening concern and each needs its own review.

The branch is kept as the source for the split; nothing is lost.

## Unit 1 — #670

`^B Route execl/execlp/execle through the exec guard` — the variadic adapter, `build.rs`, and the `guarded_*` dispatch split. 3 files, 303 changed lines. Includes a differential proof in the pinned toolchain image showing all three variadic forms escape the guard on `main` and are blocked after.

Two pieces of this branch were dropped there rather than carried: the `WORKCELL_SYMVER` block (never compiled — `build.rs` always passed `-DWORKCELL_NO_SYMVER`, and the container proof shows unversioned interposition already wins against glibc's versioned `execl@GLIBC_2.17`), and `tests/exec_variadic_probe.c` (nothing compiled it; the container harness covers the same entry points with a `main`-versus-branch control). Details in #670.

## Remaining units

Each leaves the crate building and `cargo test`/`clippy` green, and each lands a guard whole — no classifier arrives without the entry-point wiring that uses it.

2. **Bounded exec inputs.** `MAX_EXEC_*` limits, `ExecInputTooLarge`, `bounded_c_string`, `collect_cstring_array`/`c_path_string` returning `Result`, `current_process_env_entries`, `validate_path_value`, `resolve_exec_search_target`, `report_exec_input_block`/`report_with_errno`/`set_errno`, the fail-closed prologue in all eight entry points, and the two limit tests. ~550 lines. Goes first: it changes the parser signatures every later unit's entry-point code is written against.

3. **Loader-environment guard for native exec.** `token_is_loader_environment_assignment`/`_name`/`_unset_option` (the `env -S` and `env -u` handling), `env_has_unsafe_loader_override`, `loader_path_list_targets_mutable_native_exec`, `loader_args_target_mutable_native_exec`, `path_is_loader_sensitive`, `file_descriptor_is_loader_sensitive`, `should_block_loader_env_for_path`/`_for_fd`, the block message and reporter, wiring into all eight entry points, and `loader_control_options_reject_mutable_values`. ~450 lines.

4. **Require the approved guard preload.** The `launcher_common::GUARD_PRELOAD` constant and the `sanitize_env` restore, with its launcher test, land together with the guard side that depends on them: `env_has_approved_guard_preload`, `current_process_is_approved_native_launcher`, `APPROVED_CONTROL_SCRIPTS`, `path_is_approved_control_script`, `is_approved_launcher_control_transition`, `should_block_missing_guard_env`, the block message and reporter, entry-point wiring, and `guard_environment_requires_the_exact_preload`. ~350 lines. Restoring `LD_PRELOAD` in the launcher and requiring it in the guard are one change; splitting them would leave a half-wired guard in either order.

5. **Trusted-immutable provenance, path side.** `MUTABLE_EXEC_ROOTS`, `AT_EACCESS_FLAG` and the `faccessat` extern, `path_is_magic_exec_target`, `effective_access_can_write`, `path_is_trusted_immutable`, `path_ancestors_are_trusted_immutable`, `lexical_normalized_path`, `path_with_dirfd_base`, `path_has_untrusted_provenance`, `path_has_mutable_provenance`, `path_resolves_to_mutable_root`, the extended `should_block_mutable_native_exec` and its block message, wiring into the path-taking entry points, and the path-provenance tests. ~600 lines.

6. **Trusted-immutable provenance, descriptor side.** `fd_signature`, `fd_target_is_trusted_immutable`, `fd_target_is_untrusted_exec`, `file_descriptor_is_mutable_native_exec`, wiring into `execveat`/`fexecve`, and the descriptor tests (`anonymous_elf_descriptor_is_untrusted_and_blocked`, `trusted_runtime_elf_descriptor_is_pass_through`, `mutable_root_descriptor_is_not_trusted_after_path_resolution`, `deleted_elf_descriptor_is_untrusted`, `non_executable_elf_in_writable_directory_is_blocked_for_loader_use`). ~450 lines. Depends on 5 for the trust predicate.

7. **Untrusted shebang to native interpreter.** `shebang_path_is_untrusted_native_exec`, `env_command_targets_untrusted_native_exec`, `buffer_targets_untrusted_native_via_shebang`, `file_descriptor_is_untrusted_shebang_native_exec`, `path_is_untrusted_shebang_native_exec`, the chunked file comparison, wiring, and `mutable_shebang_native_interpreters_and_loader_targets_are_rejected` plus `streaming_file_comparison_seeks_and_bounds_memory`. ~400 lines.

8. **memfd snapshot execution, direct exec paths.** `MAX_MUTABLE_EXEC_SNAPSHOT_BYTES`, `MUTABLE_EXEC_COPY_CHUNK_BYTES`, `MFD_EXEC_FLAG`, `MutableExecPreparation`, `mutable_exec_preparation`, `snapshot_mutable_file`, `fd_exec_path`, `close_prepared_fd`/`_preserving_errno`, `execute_snapshot_execveat`, `execute_snapshot_execveat_with_shell_fallback`, `execute_prepared_execve` and `execute_prepared_search_execve`, rewiring `execve`/`execv`/`execvp`/`execvpe` onto them, and `mutable_snapshot_is_sealed_and_stable_after_source_rewrite` plus `mutable_script_snapshot_executes_with_a_null_argument_vector`. ~700 lines.

9. **memfd snapshot execution, descriptor and spawn paths.** `mutable_fd_preparation`, `duplicate_owned_descriptor`, `descriptor_is_shebang`, `prepare_descriptor_for_exec`, `ExecveatPreparationTarget`, `execveat_preparation_target`, `execute_prepared_execveat`, `execute_prepared_spawn`, rewiring `execveat`/`fexecve`/`posix_spawn`/`posix_spawnp`, and `mutable_spawn_with_file_actions_fails_closed_before_child_creation`, `nonempty_execveat_at_empty_path_still_prepares_the_path`, `outside_alias_into_mutable_runtime_is_blocked_before_snapshot`. ~600 lines.

Order matters for 2 before everything, 5 before 6, and 5-7 before 8-9; 3 and 4 are independent of the rest and can land whenever.

PR #662, opened 2026-09-06
Split release image build from registry handoff

omkhar:
Remove registry, signing, and attestation authority from every job that builds or assembles release content.

## Summary

- `build-amd64-image` is a new job: the amd64 runtime image is built to an OCI archive instead of pushed by digest from the publishing job. `build-arm64-image` loses its GHCR login, its `packages: write` scope, and its `release` environment gate, and now emits the same kind of archive.
- Each builder emits a signed handoff manifest (`create-release-image-handoff`) binding repository, run, tag, commit, platform, image digest, preflight manifest and config digests, and the archive SHA-256. The Go validator walks the OCI archive, rejects non-canonical, special, unexpected, and duplicate members, verifies every blob against its own name, and requires exactly one platform descriptor for the bound manifest digest.
- `bind-release-subjects` is a new read-only job that independently constructs the nine non-image signing subjects before the approval gate.
- `release` drops to `contents: read` and only assembles: it verifies both bound archives against preflight, builds the multi-arch OCI layout offline with pinned ORAS, and uploads an unsigned bundle. `docker buildx imagetools` no longer assembles the published manifest.
- `sign-release` is the only privileged job. It runs in the `release` environment, does not check out or execute repository code, validates the handoff and every byte of the bound subjects before publishing the layout, then signs and attests.
- `ValidateReleaseWorkflowAuthoritySplit` enforces the split: the four unprivileged jobs must grant only `contents: read`, the signer must hold the exact environment, permissions, and dependencies, and the whole parsed `sign-release` job must match a pinned SHA-256 contract so an added step, changed command, or changed action input fails the check.
- Threat model, workflow, releasing, and retention docs plus the lane and retention policies are updated for the new jobs and artifacts.

## Testing

- `go build ./...` — clean.
- `go vet ./...` — clean.
- `go test ./internal/metadatautil/ -count=1` — ok, covering the new handoff and authority-split suites.
- `go test ./internal/host/release/ -count=1` — ok (asset inventory, tag-policy gate, and inventory-parser mutation suites).
- `go test ./... -count=1` — all packages ok; `internal/host/c3certify` and `internal/host/launcher` each flake once under full-suite load and pass in isolation on this branch and on `main`.
- `actionlint .github/workflows/release.yml` — no findings.
- `zizmor .github/workflows/release.yml` — no findings.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,6 +32,8 @@ env:
   WORKCELL_BUILDKIT_IMAGE: moby/buildkit:buildx-stable-1@sha256:28a898719c18a33f4e8000685287fa36fd0dd9560c6440227d3a732d79bb41d8
   WORKCELL_BUILDX_VERSION: v0.37.0
   WORKCELL_COSIGN_VERSION: v3.1.3
+  WORKCELL_ORAS_VERSION: 1.3.3
+  WORKCELL_ORAS_LINUX_AMD64_SHA256: 9ce999f8d2de03fc03968b29d743077a58783e545e5eaa53917ca177352d0e59
   WORKCELL_QEMU_IMAGE: tonistiigi/binfmt:qemu-v10.2.3-68@sha256:400a4873b838d1b89194d982c45e5fb3cda4593fbfd7e08a02e76b03b21166f0
   WORKCELL_SYFT_VERSION: v1.51.1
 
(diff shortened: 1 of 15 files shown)


omkhar:
Superseded: this branch is +1334/-151 across 15 files, over the 1200-line PR shape gate. Splitting it into two dependency-ordered PRs.

- Unit 1: #666 — the handoff validator (`internal/metadatautil/release_image_handoff*.go` and tests) plus the `create-release-image-handoff` citools subcommand. 568 lines.
- Unit 2: to follow — `.github/workflows/release.yml` job split, `ValidateReleaseWorkflowAuthoritySplit` in `workflows.go`, the pinned-input release assertions, the lane and retention policies, and the threat model, workflow, releasing, and retention docs. ~917 lines.

Closing in favour of the split.

PR #674, opened 2026-09-07
Add the apt broker privileged helper runner

omkhar:
Unit 2 of the apt broker Go series that replaced #651. Unit 1 is #668 (merged): the wire protocol and client. This unit adds the privileged helper runner that the socket server will call in unit 3.

## What this adds

`internal/aptbroker/helper_common.go` (46), `helper_common_test.go` (40), `helper_linux.go` (415), `helper_linux_test.go` (418), `helper_other.go` (16). 935 lines, 5 files, 1 area, no binaries.

The runner starts the helper in its own process group and then **proves** that ownership with `getpgid` before it is willing to signal anything. Every later signal re-validates the same proof, so a recycled PID or a helper that lost its group can never be signalled as a group. Teardown is TERM, then a bounded grace on a pidfd exit watcher, then KILL plus an exact `Process.Kill()` on the known handle.

Four outcomes race for the run: helper exit, client disconnect, request timeout, output overflow. `committedHelperCause` commits exactly one winner via `context.AfterFunc`, so a helper that exits at the same instant the client hangs up cannot produce two conflicting verdicts. Output is capped at `MaxOutputBytes` per stream by a bounded buffer, and the drain is itself bounded so a helper holding its pipes open cannot stall cleanup.

`helper_other.go` is an unsupported-platform stub, so the package still builds on Darwin for local development.

## Series notes

- `runOwnedHelper` has no caller until the socket server arrives in unit 3. `scripts/check-dead-code.sh` reports it (alongside the pre-existing `internal/injectionpolicy/reader_acl.go` report) and still exits 0.
- The container keeps the shell broker until the final wiring unit: `runtime/container/apt-broker.sh` and `sudo-wrapper.sh` are untouched here, so nothing regresses at this intermediate state.
- No new module dependency: `golang.org/x/sys` is already a direct requirement in `go.mod`.

## Testing

- `go build ./...` and `GOOS=linux go build ./...` clean.
- `go vet ./...` and `GOOS=linux go vet ./internal/aptbroker/` clean. `gofmt -l internal/aptbroker/` empty.
- `go test ./internal/aptbroker/... -count=1` ok on Darwin.
- Same suite run inside a Linux container so the `//go:build linux` half actually executes: all 17 helper tests pass, plus the unit-1 protocol and client tests. This covers the teardown ordering assertions, the unproved-ownership and ownership-changed refusals, the startup-failure cleanup paths, the committed-cause race, and live helper runs for exit status, environment, output bounds, timeout status 124 and pre-cancellation (the helper must not have started).
- `scripts/check-dead-code.sh`, `scripts/check-public-repo-hygiene.sh`, `scripts/check-public-contract.sh`, `scripts/check-pr-shape.sh` all exit 0.


omkhar:
@codex review

omkhar:
Round 1 findings addressed in `fa5010da`. Both were correct; thanks.

**P2 — Preserve signal-derived helper exit statuses.** Confirmed. `ProcessState.ExitCode()` returns `-1` for a signalled child, so the old code collapsed every signal death to `1`. Verified the legacy behaviour it diverged from: `sh -c 'kill -TERM $$'` reports `143` through `wait` in the container. `exitStatus` now derives `128+N` from `syscall.WaitStatus` when `Signaled()` is set, which stays inside the protocol's 0-255 status domain. The committed-cause statuses are unaffected: a timeout still reports `124` and an output overflow still reports `1`, because `helperResponse` overrides after the fact. New test `TestExitStatusPreservesSignalTermination`.

**P2 — Retry pidfd polling when a signal interrupts it.** Confirmed at the source: `unix.Poll` in `golang.org/x/sys@v0.47.0` is a thin wrapper over `Ppoll` with no `EINTR` retry, and the Go runtime sends `SIGURG` for goroutine preemption, so this was reachable. The loop now continues on `EINTR` and reserves `watcher.err` for a non-transient failure. `watch` delegates to `watchWith(poll)` so both paths have deterministic tests: `TestHelperWatcherRetriesInterruptedPoll` and `TestHelperWatcherReportsNonTransientPollFailure`.

**Also fixed: the CI "Validate repository" failure**, which was mine and not a review finding. `unixConnectionPair` bound a socket under `t.TempDir()`, and the longest test name under the CI `TMPDIR` overran the AF_UNIX `sun_path` limit:

```
listen unix /tmp/workcell-home-1001/.tmp/TestRunOwnedHelperPublishesOwnershipBeforeCancellationCleanup.../001/socket: bind: invalid argument
```

Reproduced exactly in a container with `TMPDIR=/tmp/workcell-home-1001/.tmp`. Renaming the one test would only have moved the cliff, so the pair now comes from `syscall.Socketpair` and binds no path at all.

Evidence: whole package green in a Linux container under that same long `TMPDIR`, including `-race` and `-count=30`. `go build ./...`, `GOOS=linux go build ./...`, `go vet ./...`, `GOOS=linux go vet ./internal/aptbroker/`, `gofmt -l` all clean. `check-dead-code.sh`, `check-public-repo-hygiene.sh`, `check-pr-shape.sh` exit 0.

omkhar:
@codex review

omkhar:
@codex review

omkhar:
**Review-loop status: round 2 blocked by a Codex usage limit, not by findings.**

Round 1 reviewed `075813ab` and raised two findings, both P2 — no P0 and no P1. Both were confirmed correct, fixed in `fa5010da`, reacted to, and their threads resolved.

Two re-review triggers (03:43 and 04:00) both returned "You have reached your Codex usage limits for code reviews", so there is no clean marker naming the current head `fa5010da`.

Standing in for that, on the current head:

- All 11 non-skipped CI checks pass, including **Validate repository (17m4s)** — the check that failed on `075813ab` for the `sun_path` bug this round fixed.
- Whole `internal/aptbroker` package green in a Linux container under the CI's own long `TMPDIR`, with `-race` and `-count=30`.
- `go build ./...`, `GOOS=linux go build ./...`, `go vet ./...`, `GOOS=linux go vet ./internal/aptbroker/`, `gofmt -l` all clean.

Worth a fresh `@codex review` before merge once the usage limit resets.

omkhar:
@codex review

PR #673, opened 2026-09-07
^R Delegate Codex TOML invariants to Go (bash -n, shellcheck, shfmt, full go test pass; deletes the awk TOML parser)

omkhar:
## What

Unit 2 of the series that replaces PR #649. Unit 1 is #667 (merged): it added
`internal/metadatautil/codex_config.go` and the `workcell-citools`
`workcell-codex-toml-invariants` subcommand, additive only, with
`scripts/verify-invariants.sh` untouched.

This unit switches the bash over and deletes the superseded shell code.
One file, +12 / -659.

- The Codex TOML invariant region of `scripts/verify-invariants.sh` becomes a
  single `go_verify_citools workcell-codex-toml-invariants "${ROOT_DIR}" || exit 1`
  at the region's original position, so its order relative to the neighbouring
  checks is unchanged. The Go subcommand keeps the original check order
  internally: the routing-config check still runs between the managed-config and
  profile-layer validations.
- Deletes the awk TOML parser (`toml_section_assignments`, `toml_section_names`)
  and the `require_toml_assignment`, `require_toml_key_absent`, and
  `require_toml_exact_keys` helpers, together with
  `verify_codex_managed_config_invariants`,
  `verify_codex_profile_layer_invariants`,
  `assert_codex_managed_config_rejected`, and the in-script mutant fixtures.
  The mutant fixtures live on as Go unit tests in
  `internal/metadatautil/codex_config_test.go` (shipped in #667).
- Deletes the standalone `go_verify_citools validate-codex-routing-configs` call.
  The new subcommand runs that same check through
  `metadatautil.ValidateCodexRoutingConfigs`, so leaving the call in place would
  run it twice.

### Dead code note

The series plan named a `require_toml_section_absent` helper to delete as
zero-call-site dead code. That function does not exist on `main`, and it does not
exist on the source branch either (`git grep require_toml_section_absent
2068b98f^` returns nothing). The plan entry was a mislabel of
`require_toml_key_absent`, which does have call sites (inside
`verify_codex_managed_config_invariants` and
`verify_codex_profile_layer_invariants`) and is deleted here along with those
callers. No separate dead-code deletion was needed.

## No other caller references the deleted helpers

`grep -rn 'toml_section_assignments\|toml_section_names\|require_toml_'` over the
whole tree, `.git` excluded, before the change: 26 hits in
`scripts/verify-invariants.sh` (all in the replaced region), 4 in
`internal/metadatautil/codex_config.go`, and 1 in
[...]

PR #681, opened 2026-09-07
Batch static Go delegations into one citools call

omkhar:
Unit 3 of the #649 series. Units 1 and 2 are #667 (Go TOML port + `workcell-codex-toml-invariants` subcommand) and #673 (bash delegation, awk TOML parser deleted); both are merged. The remaining work is the live-lane extraction unit(s).

## What this does

Adds the `workcell-citools workcell-check-batch` subcommand and points the eight contiguous runs of static delegations in `scripts/verify-invariants.sh` at it.

`workcell-check-batch ROOT_DIR CHECK[=ARG] [CHECK...]` runs the subcommand table's single-argument checks in argv order inside one process. A `ROOT_DIR` check is named bare and receives the shared root; a `SETTINGS_PATH` check is spelled `CHECK=PATH`. It stops at the first failure and returns that failure's error unchanged, so `die()` prints stderr byte-identical to running the failing check's own subcommand and exits 1 the same way. An unknown check, a non-batchable subcommand, a `ROOT_DIR` check given `=ARG`, or a `SETTINGS_PATH` check missing `=ARG` is a usage error (exit 2 via `dieUsage`), matching `main`'s dispatch for a malformed invocation.

48 static delegation call sites become 19: 37 sites collapse into 8 batch calls, and the per-check `go run` spawn count drops with them.

## Ordering parity

The only permitted behaviour change is per-check calls collapsing into batch calls covering the same checks in the same first-failure order. No batch straddles an inline shell check.

The claude settings `for settings_path` loop is unrolled into four batch entries in the original loop order: `settings.json` then `managed-settings.json`, and within each file `mcp-project-servers` then `guard-bash-hook`.

Evidence: the ordered citools check sequence was extracted from the script before and after the change (batch argv lists expanded, the settings loop unrolled). Both sequences are 50 entries and `diff` between them is empty.

## Comment collapse

The multi-line D3 banners ahead of the batched calls collapse to one line each pointing at the owning `CheckXxx`, because the parity notes those banners duplicated already live in `internal/workcellhardening`. The stays-inline notes are kept verbatim. The shared `|| exit 1` ERR-trap rationale, previously repeated in nearly every banner, now lives once above `go_verify_citools`.

`internal/workcellhardening`'s package doc gains the batched-delegation contract, and its claude-settings note now says "once per claude settings file, in the script's original per-file order" instead of describing the deleted shell for-loop.

## Testing

- `go build ./...`, `gofmt -l`, `go vet ./cmd/workcell-citools/ ./internal/workcellhardening/` clean.
- `go test ./cmd/workcell-citools/ ./internal/metadatautil/ -count=1` ok. Also `./internal/workcellhardening/ ./internal/testkit/ -count=1` ok (70.9s).
- New tests: first-failure ordering with byte-identical stderr against an individually run subcommand (both orderings of a two-check batch, proving whichever check is listed first owns the batch's entire stderr), the mixed `SETTINGS_PATH` + `ROOT_DIR` success form against the real repo, and the five usage-error invocations.
- `bash -n`, `shellcheck -x`, and `shfmt -ln=bash -i 2 -ci -d` on `scripts/verify-invariants.sh` all clean.
- All eight rewritten batches executed directly against this repo: every one exits 0.
- Static lane: `scripts/verify-invariants.sh` truncated immediately after the sixth batch call ran to completion with exit 0 in a standard clone, covering 32 of the 50 checks in the real script flow together with every inline check ahead of them. The last two batches sit behind the Darwin live-Colima lanes, so they were exercised through the subcommand directly rather than by starting VMs.
- `./scripts/check-dead-code.sh` exit 0 (its one report, `internal/injectionpolicy/reader_acl.go:10 isPolicyACLName`, predates this branch).

## Shape

4 files, +252/-217 = 469 changed lines.

### DIFF
--- a/cmd/workcell-citools/main.go
+++ b/cmd/workcell-citools/main.go
@@ -175,6 +175,7 @@ func subcommands() []subcommand {
 		{"workcell-hostgate-entrypoint-sanitize", "ROOT_DIR", 1, 1, hardeningCheck(workcellhardening.CheckHostGateEntrypointSanitize)},
 		{"workcell-precommit-upstream-pin-gate", "ROOT_DIR", 1, 1, hardeningCheck(workcellhardening.CheckPrecommitUpstreamPinGate)},
 		{"workcell-trusted-docker-client-rg", "ROOT_DIR", 1, 1, hardeningCheck(workcellhardening.CheckTrustedDockerClientRg)},
+		{"workcell-check-batch", "ROOT_DIR CHECK[=ARG] [CHECK...]", 2, -1, cmdWorkcellCheckBatch},
 	}
 }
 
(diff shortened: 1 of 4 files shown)


omkhar:
@codex review

omkhar:
@codex review

omkhar:
@codex review

omkhar:
@codex review

omkhar:
@codex review

PR #702, opened 2026-09-08
.R Move the rest of the live Colima lane into verify-invariants-live.sh

omkhar:
Unit 4b, the final unit of the series that replaces #649. Units 1-3 are #667,
#673 and #681; unit 4a is #696. All merged.

## What is left to move

Unit 4a moved the two only-caller helpers and the live lane up to the
detached-session attach handshake into `scripts/verify-invariants-live.sh`, and
gave the fragment its own copy of the seven-line Darwin and free-space gate so
that a sourced file parses on its own. The parent kept a second copy of that
gate for the half of the lane it still held.

This unit moves that remaining half: the `session send` assertions, the
detached-session monitor finalization checks, the live-debug audit lane, and
the strict-refresh profile checks. The parent's duplicate gate goes with it,
because the fragment already carries the one gate the whole lane needs. What
remains in the parent at that position is the single `source` line unit 4a put
there.

`scripts/verify-invariants.sh` drops from 7594 to 7198 lines.

## The awk ordering assertion

`tests/scenarios/shared/test-session-commands.sh` asserts that the detached
monitor wait runs before the audit-directory and monitor-state-file absence
checks:

```
  index($0, "if ! wait_for_detached_session_monitor_exit") { wait_line = NR }
  index($0, "test ! -e \"${DETACHED_SESSION_AUDIT_DIR}\"")  { audit_line = NR }
  index($0, "test ! -e \"${DETACHED_SESSION_MONITOR_STATE_FILE}\"") { state_line = NR }
```

All three anchors are inside this unit's region. Unit 4a left the assertion
pointing at the parent for exactly this reason; here it repoints at the
fragment. Against the fragment it resolves to `wait=530 audit=535 state=536`
and passes.

## Move proof

Against `db445829` (`origin/main` at branch point):

```
$ diff <(git show db445829:scripts/verify-invariants.sh | sed -n '6299,6684p') \
       <(sed -n '423,808p' scripts/verify-invariants-live.sh)
$ diff <(git show db445829:scripts/verify-invariants.sh | sed '6291,6686d') \
       scripts/verify-invariants.sh
```

Both empty. The first is the moved region against its new location; the second
is the parent against `main` minus exactly the deleted span. The fragment is
also reproducible from `main`: splicing the extracted region into
`db445829:scripts/verify-invariants-live.sh` ahead of its final two lines
reproduces the new fragment byte for byte.

The deleted span is 396 lines: one blank line, the seven-line gate, the 386
moved lines, the `fi` that closed the free-space `else`, and the `fi` that
closed the Darwin gate. Only the last four of those are not a move; the
fragment already provides their equivalents.

Ordered `go_verify_citools` call list, taken in execution order (parent up to
the `source` line, then the whole fragment, then the parent remainder) with
`grep -oE '^[[:space:]]*go_verify_citools[[:space:]]+[a-z0-9-]+' | awk '{print $NF}'`:
20 entries before, 20 entries after, diff empty. One of them,
`workcell-smoke-apt-broker-probe`, is inside the moved region and keeps its
position.

## Testing

- `bash -n` on both scripts and on `test-session-commands.sh` - exit 0
- `shellcheck -x` on all three - no findings
- `shfmt -ln=bash -i 2 -ci -d` on all three - no diff
- `WORKCELL_VALIDATE_REPO_PROFILE=repo-core ./scripts/validate-repo.sh` - exit 0.
  Its scenario lane runs `shared/session-commands`, which is where the repointed
  awk assertion lives (`Results: passed=17 failed=0 skipped=5`).
- `go test ./internal/testkit/ ./internal/metadatautil/ ./internal/workcellhardening/
  ./cmd/workcell-hostutil/ ./cmd/workcell-citools/ -count=1` - all ok.
  `testkit` and `workcell-hostutil` assert on `verify-invariants.sh` contents.
- **A full `./scripts/verify-invariants.sh` A/B on a Darwin host, into two
  standard clones, both exit 0.** This host has `/opt/homebrew/bin/colima` and
  `/opt/homebrew/bin/docker`, so the Darwin gate is true and the live lane ran
  end to end on both sides - managed Colima profile starts, the live-debug audit
  lane, the detached session lane, and the strict-refresh checks. So the moved
  region was executed, not only parsed. Normalizing profile PIDs, tmpdir names,
  workspace paths, timestamps and the checked-out SHA, the two runs produce
  identical output, 145 lines each.
  Sequence: baseline on plain `main`, then the branch content, both exit 0. An
  earlier A/B pair on the `23696d68` base gave the same result, with two extra
  `go: downloading` lines on the cold-cache side.
- Review loop deferred.

Shape: 3 files, +387/-397 = 784 changed lines, 2 areas.

## Series

`codex/invariants-toml-port-v2` is fully extracted once this merges.


omkhar:
@codex review

PR #709, opened 2026-09-08
Add the shell portability banlist (S14)

omkhar:
## Summary

Adds a banlist for the GNU-only and Bash-4-only constructs that fail on the
host baseline. `scripts/check-doc-links.sh:5` already declares the contract
("Kept bash-3.2 compatible ... because the host baseline is macOS /bin/bash
3.2"), and nothing enforced it. Codex review history shows 29 accepted
findings of this class before pull request 600.

## Every rule was measured, not assumed

Measured on macOS 26.6.2 with the system path `/usr/bin:/bin` and the system
`/bin/bash` 3.2.57:

| Construct | Result | On the list |
|---|---|---|
| `mapfile`, `readarray`, `wait -n`, `declare -A`, `${x^^}` | fails on bash 3.2 | yes |
| `stat -c` | fails | yes |
| `date -d` | fails | yes |
| `grep -P` | fails | yes |
| `find -printf` | fails | yes |
| `sed -i` without a suffix | fails | yes |
| `sha256sum` | fails on `/usr/bin:/bin` | yes |
| `readlink -f` | **works** | no |
| `mktemp` and `mktemp -d` without a template | **works** | no |

Measurement removed two candidates the report proposed. A `mktemp` rule alone
would have produced 32 findings that are not defects on this baseline.

Honest caveat on `sha256sum`: macOS 26 ships `/sbin/sha256sum`, and the
repository's trusted path includes `/sbin`, so the command resolves there
today. The rule stays because four accepted review findings name it and
because `scripts/verify-release-artifact.sh:102` already carries a `shasum`
fallback for a host that lacks it.

## Three discriminators

1. **The runtime tree is out of scope.** `runtime/**` only runs inside the
   Debian image. Its scripts are also digest-bound by
   `runtime/container/control-plane-manifest.json`, so a comment added to one
   of them is a manifest change.
2. **A heredoc body is skipped.** It is text the script hands to another
   interpreter, which is bash 5 inside the image for every such body in the
   tree. This is the reviewer's own class-D lesson: read the commands the
   script runs, not the text it carries. It clears 8 `stat -c` lines in
   `scripts/container-smoke.sh` that run inside a container.
3. **A modern-bash assertion clears the bash-version rules.** A script that
   calls `workcell_require_modern_privileged_bash` re-executes itself under
   bash 4, which is why `scripts/validate-repo.sh` may use `declare -A`. Five
   scripts do this. The userland rules still apply to them.

## Repairs, not exemptions

Two host-side scripts are repaired:

- `tests/scenarios/shared/test-aws-ec2-ssm-launch-smoke.sh:72`: `mapfile`
  becomes a `while read` loop.
- `tests/scenarios/shared/test-publish-github-release.sh:107`:
  `"${commit_sha^^}"` becomes `tr '[:lower:]' '[:upper:]'`.

`scripts/ci/run-docs-in-validator.sh:68` was flagged for review rather than
exemption. Review result: the `mapfile` sits inside the `bash -lc` command
string that `docker run` hands to the validator image, so it runs on bash 5
inside the image. It carries an inline exemption naming that.

The `runtime/container/apt-broker.sh` item from the report is gone: pull
request 699 deleted that file.

Sixteen other reviewed lines carry an inline exemption with its reason:
the BSD form is tried first (3), the shasum fallback follows (1),
`require_tool sha256sum` fails closed (3), the Linux CI or release lane (8),
and the validator image command string (1).

## Not done, with the reason

The report also asked for a `BASH_VERSINFO[0] -ge 3` assertion in the shared
library. Skipped: every bash is 3 or newer, so the assertion can never fire.
The repository already carries the assertion that matters,
`workcell_require_modern_privileged_bash`, and this check now reads it.

## Follow-up found while measuring

`scripts/generate-control-plane-manifest.sh` declares
`# generated-artifact: none` after pull request 707, but it also writes the
tracked `runtime/container/control-plane-manifest.json`. Its declaration
should name that file so the freshness gate covers it. Not changed here to
keep this change to one subject.

## Evidence

- `WORKCELL_VALIDATE_REPO_PROFILE=repo-core ./scripts/validate-repo.sh` passes.
- `./scripts/check-shell-portability.sh` passes.
- `./scripts/check-generated-artifacts.sh` passes.
- `./scripts/check-doc-language.sh` passes.
- `./scripts/check-validator-anchoring.sh` passes.
- `./scripts/verify-control-plane-manifest.sh` passes.
- `bash -n` passes on every edited script.
- `./scripts/check-pr-shape.sh`: files=23 lines=429 areas=5.

Review loop deferred.

### DIFF
--- a/cmd/workcell-citools/main.go
+++ b/cmd/workcell-citools/main.go
@@ -110,6 +110,7 @@ func subcommands() []subcommand {
 		{"check-validator-anchoring", "REPO_ROOT", 1, 1, cmdCheckValidatorAnchoring},
 		{"check-doc-language", "REPO_ROOT", 1, 1, cmdCheckDocLanguage},
 		{"check-generated-artifacts", "REPO_ROOT", 1, 1, cmdCheckGeneratedArtifacts},
+		{"check-shell-portability", "REPO_ROOT", 1, 1, cmdCheckShellPortability},
 		{"verify-reproducible-build", "OCI_EXPORT_A OCI_EXPORT_B REPRO_PLATFORMS REPRO_MANIFEST_PATH SOURCE_DATE_EPOCH", 5, 5, cmdVerifyReproducibleBuild},
 		{"generate-reproducible-build-manifest", "OCI_EXPORT REPRO_PLATFORMS OUTPUT_PATH SOURCE_DATE_EPOCH", 4, 4, cmdGenerateReproducibleBuildManifest},
 		{"verify-reproducible-build-manifest", "OCI_EXPORT REPRO_PLATFORMS MANIFEST_PATH", 3, 3, cmdVerifyReproducibleBuildManifest},
@@ -541,6 +542,10 @@ func cmdCheckDocLanguage(args []string) error {
 
 func cmdCheckGeneratedArtifacts(args []string) error {
 	return metadatautil.CheckGeneratedArtifacts(args[0])
+}
+
+func cmdCheckShellPortability(args []string) error {
+	return metadatautil.CheckShellPortability(args[0])
 }
 
 func cmdCheckPinnedInputs(args []string) error {
--- a/internal/metadatautil/portability_check.go
+++ b/internal/metadatautil/portability_check.go
@@ -1,0 +1,193 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+package metadatautil
+
+import (
+	"errors"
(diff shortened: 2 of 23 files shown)


omkhar:
CI status: every required check passes. The `Hostile environment` axes `tmpdir`, `uidmap` and `root` fail on `origin/main` today and are not related to this change; see the same note on pull request 704.

Review loop deferred.

PR #710, opened 2026-09-08
Add the shared hostile-field corpus (S15)

omkhar:
## Summary

Adds `internal/testkit/hostilefields.go`: one exported table of untrusted
field values, plus two drivers over it, plus a fuzz target over the audit
path encoder.

Codex review history shows 38 accepted findings of untrusted-field encoding
into structured records across 20 pull requests, all accepted, none rebutted.
Coverage before this change was one decoy string per package, for example
`internal/applecontainer/audit_test.go:90`.

## The corpus

25 rows, each recording the defect it reproduces: an embedded newline,
carriage return or NUL; a separator inside a value; an event sentinel spelled
inside a path; a value that is the prefix of a sentinel; a space, tab or comma;
a double-quoted, single-quoted or backslash-escaped value; a literal
backslash; an octal escape; an ANSI-C escape; a bare and an already-encoded
percent sign; a non-UTF-8 byte; DEL and a high byte; a value that spells an
option; an empty and a whitespace-only value; an unmapped key name; and a long
value.

## The drivers

- `RequireRoundTripOrReject(t, encode, decode, corpus)`. Each row must survive
  the pair byte for byte, or be rejected by the encoder. Silent mangling
  fails, because a value that reads as clean while its evidence is gone is the
  defect.
- `RequireEncodedFieldIsInert(t, encode, separators, corpus)`. The encoded
  form must hold no field separator. A round trip alone does not prove this:
  an encoder that returns its input unchanged round-trips perfectly and still
  lets a newline forge a record.

## Two answers, one table

| Package | Subject | Answer |
|---|---|---|
| `internal/applecontainer` | `encodeAuditPathValue` and `decodeAuditPathValue` | encodes every row and recovers it exactly |
| `internal/shellproto` | `WriteField` and the bash reader's own split | rejects the record separators, round-trips the rest |

Both answers are correct under the contract, which is why the driver accepts
either.

## Fuzz target

`FuzzAuditPathValueRoundTrip` carries both audit-field invariants past the
corpus and seeds itself from it. Registered in
`scripts/ci/run-fuzz-in-validator.sh` and in `docs/fuzzing.md`.

A 25-second local run reached 3.4 million executions with no failure, from 25
seeds to 53 interesting inputs.

## Mutation row

`internal/mutation/runner.go` gains one row that narrows the encoder from
`c <= 0x20` to `c <= 0x1f`, which stops encoding a space and leaves a field
separator in the encoded value. `TestAuditPathValueEncodingIsInert` kills it.
`policy/mutation-score-policy.json` rises from 28 to 29 mutants.

## Evidence

- `WORKCELL_VALIDATE_REPO_PROFILE=repo-core ./scripts/validate-repo.sh` passes,
  including the mutation lane at 29 of 29 killed.
- `go test ./internal/applecontainer -fuzz FuzzAuditPathValueRoundTrip
  -fuzztime=25s` passes.
- `./scripts/check-doc-language.sh`, `./scripts/check-shell-portability.sh`,
  `./scripts/check-generated-artifacts.sh`,
  `./scripts/check-validator-anchoring.sh` and
  `./scripts/check-doc-links.sh` pass.
- `./scripts/check-pr-shape.sh`: files=7 lines=226 areas=4.

Review loop deferred.

### DIFF
--- a/docs/fuzzing.md
+++ b/docs/fuzzing.md
@@ -16,6 +16,7 @@ harness invariant.
 | `internal/tomlsubset` | `FuzzParse` | Policy TOML subset |
 | `internal/injection` | `FuzzIsAllowedSystemSymlink` | Direct-mount source chains |
 | `internal/injection` | `FuzzParseSSHDirective` | SSH configuration directives |
+| `internal/applecontainer` | `FuzzAuditPathValueRoundTrip` | Untrusted audit field values |
 
 The seed corpus includes repository configuration and invalid forms. The
 `go test ./...` pull-request lane replays each saved seed.
--- a/internal/applecontainer/hostilefields_test.go
+++ b/internal/applecontainer/hostilefields_test.go
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Omkhar Arasaratnam
+
+package applecontainer
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/omkhar/workcell/internal/testkit"
+)
+
+// auditFieldSeparators lists the bytes that end one field of an audit line.
+// An equals sign is not one of them: the reader splits a field at its first
+// equals sign, so a later one stays inside the value.
+const auditFieldSeparators = " \t\n\r\x00"
+
+func encodeAuditPathField(value string) (string, error) {
+	return encodeAuditPathValue(value), nil
+}
+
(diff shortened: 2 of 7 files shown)


omkhar:
CI status: every required check passes. The `Hostile environment` axes `tmpdir`, `uidmap` and `root` fail on `origin/main` today and are not related to this change; see the same note on pull request 704.

Review loop deferred.

PR #712, opened 2026-09-08
^B Classify the interpreter a shebang names before dispatching on what it is called, and stop the guard reading anything it cannot execute (six differential and mutation controls; user-visible execution boundary)

omkhar:
Retroactive Codex sweep over the merged exec-hardening series (#703, #705,
#706, #708). Six findings reproduced against origin/main and fixed here.

**A shebang interpreter was recognised by basename before it was classified.**
`buffer_targets_untrusted_native_via_shebang` dispatched on `env` and on
`ld-linux-*` before asking the interpreter's own provenance, and both walks
then answered for the trusted target the interpreter named. A caller-planted
`/tmp/env` in `#!/tmp/env /bin/sh` was never classified at all, and the kernel
launched it. Confirmed by probe before the fix. The provenance test moves ahead
of both dispatches, in the buffer reader and in the env walk; `/usr/bin/env`
and the real loader are trusted, so both walks still run for the image's own
lines.

**A classifier could block forever on a descriptor it could not execute.** The
shebang readers reopened the target through `/proc/self/fd` with no type test,
so `fexecve` on a pipe read end with a live writer, or `execve("/dev/stdin")`,
blocked in `read` inside the interposed call; a FIFO named by path blocked in
`open`. The kernel refuses a non-regular exec target at once. All five readers
now open through `open_regular_for_classification`, which sets `O_NONBLOCK` and
requires a regular file. `path_is_mutable_native_exec` keeps its fail-closed
answer for a regular file it cannot read, and returns the kernel's own errno
for a target proven not to be one.

**A pinned descriptor was returned without being asked the refusal questions.**
Every classifier ahead of `mutable_exec_preparation` reads the pathname, and a
relative or magic name need not still resolve to what they read. A trusted
target reached through such a name is now re-tested for the protected runtime,
a mutable shebang into it, and a loader environment, and a dynamic loader
reached that way is refused outright because its own target lives in argv,
which the preparation does not carry.

**A snapshot was classified before it was sealed**, leaving a window in which
the writable memfd was reachable through `/proc/self/fd`. The seal now lands
between the copy and the classification; `F_SEAL_WRITE` still permits the read.

**A descriptor was classified before it was pinned.** `mutable_fd_preparation`
read the caller's descriptor number, then duplicated it. The duplicate is taken
first, and both the classification and the snapshot read it.

**A snapshot copied its source off a `noexec` mount.** The memfd is
exec-capable, so a mode-0755 script under the profile's `noexec` `/tmp` gained
execution the kernel had refused. `snapshot_mutable_file` refuses a source
whose mount reports `ST_NOEXEC`, and fails closed where it cannot ask.

**The ENOEXEC shell fallback closed the descriptor it named.** A non-shebang
target keeps `FD_CLOEXEC`, so the successful `execve("/bin/sh")` closed the fd
before the shell could open the `/proc/self/fd` argument it was given.

Validation

- Host: `cargo fmt --all --check`, `cargo clippy --all-targets --offline
  --locked -- -D warnings` (zero warnings), `cargo test --offline --locked`
  29 + 17 + 1 pass.
- Pinned toolchain image (`rust:1.97.1-slim-trixie`, the Dockerfile digest,
  aarch64): `cargo build --release --locked --offline` links; `cargo test
  --offline --locked` **48 + 17 + 1** pass as root and as `--user 1001:1001`.
- The two mount-dependent controls assert rather than skip under
  `--privileged`; both print a skip reason where the premise does not hold.
- Negative controls, each fix reverted on its own in a copy:
  - interpreter classified after the dispatch: `untrusted_shebang_interpreters_are_refused` fails.
  - snapshot ignores a `noexec` source: `a_snapshot_is_refused_for_a_source_on_a_noexec_mount` fails.
  - `(dev, ino)` comparison always agrees: `a_replaced_pathname_no_longer_matches_its_descriptor` fails.
  - pinned descriptor not reclassified: `exec_preparation_pins_only_what_it_has_to` fails on the loader row.
  - readers reopen a pipe: `a_classifier_never_reads_a_pipe_or_a_fifo` never returns.
- `go build ./...`; `go test ./internal/workcellhardening/ ./internal/testkit/
  ./internal/metadatautil/ -count=1` ok.
- `./scripts/check-validator-anchoring.sh` exit 0.
- `WORKCELL_VALIDATE_REPO_PROFILE=repo-core ./scripts/validate-repo.sh` exit 0.

New tests: `a_classifier_never_reads_a_pipe_or_a_fifo`,
`a_snapshot_is_refused_for_a_source_on_a_noexec_mount`, the bind-mount
signature control in `a_replaced_pathname_no_longer_matches_its_descriptor`,
and the pinned-descriptor rows in `exec_preparation_pins_only_what_it_has_to`.

### DIFF
--- a/runtime/container/rust/src/lib.rs
+++ b/runtime/container/rust/src/lib.rs
@@ -494,6 +494,32 @@ fn duplicate_fd_file(fd: c_int) -> Option<File> {
     File::open(proc_fd_path(fd)).ok()
 }
 
+/// Opens a target to read the bytes a classifier decides on.
+///
+/// Only a regular file is executable at all, so nothing else is worth reading
+/// -- and reading one is a hazard rather than a wasted call. A pipe or a socket
+/// reopened through `/proc/self/fd` gives a description whose `read` blocks
+/// until a writer moves, so the guard would hang inside the libc call it
+/// interposed on, where the kernel would have refused the exec at once; a read
+/// that did return would take those bytes out of somebody else's stream.
+/// `O_NONBLOCK` keeps the open itself from blocking on a FIFO named by path,
+/// and has no effect on the regular files this does accept.
+fn open_regular_for_classification(path: &str) -> Option<File> {
+    use std::os::unix::fs::OpenOptionsExt;
+
+    let file = std::fs::OpenOptions::new()
+        .read(true)
+        .custom_flags(libc::O_NONBLOCK)
+        .open(path)
+        .ok()?;
+    file.metadata().ok()?.is_file().then_some(file)
+}
+
+/// The descriptor form of `open_regular_for_classification`.
+fn duplicate_regular_fd_file(fd: c_int) -> Option<File> {
+    open_regular_for_classification(&proc_fd_path(fd))
+}
+
 /// The host build has no `/proc` and never runs the guard, so it keeps the
 /// root-list answer the descriptor classifiers used before provenance.
 #[cfg(not(target_os = "linux"))]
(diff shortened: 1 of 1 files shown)


omkhar:
@codex review

PR #716, opened 2026-09-08
Add the hardened filesystem helpers (S11, part 2)

omkhar:
## Summary

Pull request 711 added the ratchet that forbids new raw `os` pathname
references in eleven trust-boundary packages. This adds the three
`internal/rootio` helpers the review corpus asks for, so the 196 baselined
references have somewhere to go.

## The helpers

**`RequireSingleLinkedRegular(info *unix.Stat_t, label, name string) error`**
Rejects a leaf that is not a single-linked regular file. A symlink, a FIFO, a
socket and a device node all pass a bare existence check, and a hard link to a
file outside the trusted root passes a file-type check as well, so the link
count is the only thing that separates the intended inode from a copy of it.
Findings: 436 ("Reject non-regular audit log leaves", "Reject hard-linked
audit logs"), 447 ("Require single-linked audit logs before export").

**Dedupe.** `internal/applecontainer/conformance.go` held the only copy of
this rule; `requireNoSymlink` becomes `requireSingleLinkedRegular` and calls
the helper. The `wantDir` parameter went with it: its single caller always
passed `false`.

**`MkdirAllSyncedAt(parent *os.File, relative string, mode os.FileMode) error`**
Opens each component relative to the previous descriptor with `O_NOFOLLOW`, so
a symlink swapped into the middle of the path is refused rather than followed,
and fsyncs the directory that gained each entry rather than only the parent of
the last one. Findings: 474 ("Fsync every ancestor created by MkdirAll",
"Fsync nested ancestors on existing-dir races", "Fsync the parent when
creating the signing dir").

The fsync runs whether this call created the component or found it present,
which is what "Fsync nested ancestors on existing-dir races" asks for: a
racing creator can return from its own mkdir before its fsync completes, or
fail it, so the loser cannot treat an existing directory as already durable.

**`StageAndPublishAt` and `StageAndCreateAt`**
One staged sibling created with `O_EXCL` under a random name, mode set through
the descriptor, contents synced, published, parent directory synced. Findings:
437 ("Sync rewritten records before renaming", "Stage records before
publishing the final path", "Randomize rewrite temp names and retry
collisions", "Preserve create-once semantics during publish", "Chmod staged
session records before rename").

`WriteFileAtomicAtNoFollow` is now the replace form under its original name,
so one implementation serves both rather than a near-identical twin. Its
fourteen existing call sites gain the content sync and the parent sync, which
is the durability half of finding 437.

The create-once form publishes with `linkat`, which refuses a name that
appeared after an absence check where `renameat` would silently replace it.
Its link-and-unlink window leaves two links for an instant, which a concurrent
`RequireSingleLinkedRegular` would refuse; that is recorded in a `ponytail:`
comment naming `renameat2` with `RENAME_NOREPLACE` as the upgrade and macOS as
the reason it is not used.

## Two reverts worth reading

I first made `ReadFileAtNoFollow` apply `RequireSingleLinkedRegular` to every
leaf it opens. Two existing tests failed and showed the change was wrong:
`internal/runtimeutil` asserts that `RewriteBundleCredentialOverride`
**succeeds** on a hard-linked bundle manifest and that the atomic replace
leaves the outside inode untouched. Hard links are accepted there by design;
the protection is the atomic replace, not rejection. I reverted it.

I also first routed the `sessions dir entry` check in `conformance.go` through
the helper, which would have added a link-count rule to a deliberately
type-only check. Reverted for the same reason. The dedupe is now exactly the
one rule that was duplicated, with no behaviour change to either site.

## Follow-up folded in

`scripts/generate-control-plane-manifest.sh` declared
`# generated-artifact: none` while it also writes the tracked
`runtime/container/control-plane-manifest.json`. It now names that file, so
the freshness gate from pull request 707 regenerates and byte-compares it on
every validate lane instead of skipping it.

## Tests and mutation

`internal/rootio/hardened_test.go` covers the five leaf shapes (regular, hard
link, symlink, FIFO, directory) and a nil stat; `MkdirAllSyncedAt` over one
component, nested components, an existing tree, an absolute path, a parent
escape, an empty path, a bare dot, and a symlinked component that must not be
followed; publication for replace, create-once refusal, mode, staged-sibling
cleanup, unsafe names, and that every published file passes the reader's own
single-link rule.

Three `internal/mutation/runner.go` rows, one per helper:
the link-count requirement, the `O_NOFOLLOW` on directory components, and the
create-once branch. `policy/mutation-score-policy.json` rises from 30 to 33.

## Evidence

- `WORKCELL_VALIDATE_REPO_PROFILE=repo-core ./scripts/validate-repo.sh` passes,
  including the mutation lane at 33 of 33.
- `./scripts/check-hardened-fs.sh`, `./scripts/check-generated-artifacts.sh`,
  `./scripts/check-doc-language.sh`, `./scripts/check-shell-portability.sh`,
  `./scripts/check-validator-anchoring.sh` and
  `./verify/invariants/control-plane-lockstep.sh` pass.
- `./scripts/check-pr-shape.sh`: files=7 lines=610 areas=3.

### DIFF
--- a/internal/applecontainer/conformance.go
+++ b/internal/applecontainer/conformance.go
@@ -18,29 +18,21 @@ import (
 
 	"github.com/omkhar/workcell/internal/host/sessions"
 	"github.com/omkhar/workcell/internal/providerid"
+	"github.com/omkhar/workcell/internal/rootio"
 )
 
-// requireNoSymlink verifies path (under stateRoot) is reached without traversing any symlink and
-// is the expected kind (regular file, or directory when wantDir), via the production no-follow
-// stat (statPathSafe: parents O_NOFOLLOW, leaf Fstatat AT_SYMLINK_NOFOLLOW). A regular leaf also
-// requires Nlink==1 (mirroring readFileSafe), so a symlink OR a hardlink to a decoy outside
-// StateRoot is rejected before any content is read.
-func requireNoSymlink(stateRoot, path, label string, wantDir bool) error {
+// requireSingleLinkedRegular verifies path (under stateRoot) is reached without traversing any
+// symlink and is a single-linked regular file, via the production no-follow stat (statPathSafe:
+// parents O_NOFOLLOW, leaf Fstatat AT_SYMLINK_NOFOLLOW). The type and link-count rule itself is
+// rootio.RequireSingleLinkedRegular, which the hardened readers apply to every leaf they open, so
+// a symlink, a FIFO or a hardlink to a decoy outside StateRoot is rejected by one rule rather than
+// by a copy of it.
+func requireSingleLinkedRegular(stateRoot, path, label string) error {
 	st, err := statPathSafe(stateRoot, path)
 	if err != nil {
 		return fmt.Errorf("%s %q: %w", label, path, err)
 	}
-	want := unix.S_IFREG
-	if wantDir {
-		want = unix.S_IFDIR
-	}
-	if int(st.Mode)&unix.S_IFMT != want {
-		return fmt.Errorf("%s %q is not the expected file type (mode %#o) — a symlink/decoy is rejected", label, path, st.Mode&unix.S_IFMT)
-	}
-	if !wantDir && st.Nlink != 1 {
-		return fmt.Errorf("%s %q is multiply linked (%d links) — a hardlink to a decoy is rejected", label, path, st.Nlink)
-	}
-	return nil
+	return rootio.RequireSingleLinkedRegular(&st, label, path)
 }
 
 // canonicalLayout is the set of on-disk paths derived purely from the state root + contract + case
@@ -229,7 +221,7 @@ func RunConformance(ctx context.Context, target ConformanceTarget, contract Cont
 	}
 	// The export reads the audit log by FOLLOWING record.AuditLogPath (== layout.auditLog); reject a
 	// symlink/hardlink there first, else a decoy log outside StateRoot could be exported.
-	if err := requireNoSymlink(c.StateRoot, layout.auditLog, "audit log", false); err != nil {
+	if err := requireSingleLinkedRegular(c.StateRoot, layout.auditLog, "audit log"); err != nil {
 		return ConformanceResult{}, err
 	}
 	exported, err := sessions.ExportSessionRecordInRoots([]string{c.StateRoot}, c.SessionID)
--- a/internal/applecontainer/conformance.go
+++ b/internal/applecontainer/conformance.go
@@ -24,9 +24,11 @@ import (
 // requireSingleLinkedRegular verifies path (under stateRoot) is reached without traversing any
 // symlink and is a single-linked regular file, via the production no-follow stat (statPathSafe:
 // parents O_NOFOLLOW, leaf Fstatat AT_SYMLINK_NOFOLLOW). The type and link-count rule itself is
-// rootio.RequireSingleLinkedRegular, which the hardened readers apply to every leaf they open, so
-// a symlink, a FIFO or a hardlink to a decoy outside StateRoot is rejected by one rule rather than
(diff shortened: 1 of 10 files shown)


omkhar:
@codex review

omkhar:
Round 1 fix pushed as `0661d1a6`. The one finding is accepted and reacted to.

**Remove the false all-readers enforcement claim.** Correct, and it was a real trap. The comment claimed the hardened readers apply the single-link rule to every leaf; they do not. `ReadFileAtNoFollow` checks the file type only, because `internal/runtimeutil` reads a hard-linked bundle manifest on purpose and relies on the atomic replace rather than on rejection. The comment now states that limit and says a caller needing the link-count guarantee calls the rule itself.

This is the same boundary the pull request body records under "Two reverts worth reading": an earlier draft did apply the rule in the readers, and two existing tests proved that wrong.

Evidence: `go test ./internal/applecontainer` passes; `./scripts/check-hardened-fs.sh`, `./scripts/check-generated-artifacts.sh`, `./scripts/check-doc-language.sh` and `./scripts/check-validator-anchoring.sh` pass.

omkhar:
@codex review

omkhar:
Round 2 fixes pushed as `69a5ecbe`. Both findings accepted; each reacted with a positive reaction.

1. **Make create-once publication atomic.** The reviewer named the right API and it exists. The kernel now refuses an existing name in the same step that publishes: `renameatx_np` with `RENAME_EXCL` on Darwin, `renameat2` with `RENAME_NOREPLACE` on Linux, in `rename_darwin.go` / `rename_linux.go` / `rename_other.go`.

   Verified on macOS 26 against a real temporary directory: a fresh name publishes with a nil error, and an existing name returns `EEXIST`. The link-and-unlink path is therefore unreachable on the host baseline. It stays as a fallback for a kernel or filesystem that refuses the one-step form (`ENOTSUP`, `ENOSYS`, `EINVAL`), keeps its `ponytail:` ceiling scoped to that path only, and now reports a failed removal through a sentinel so the caller neither unpublishes the name nor skips the parent fsync — the second half of the finding.

2. **Keep the duplicated directory descriptor close-on-exec.** `MkdirAllSyncedAt` takes its duplicate with `F_DUPFD_CLOEXEC` instead of `dup`.

The `StageAndCreateAt` doc comment no longer describes `linkat` as the mechanism, since it is now the fallback rather than the path.

Evidence: `WORKCELL_VALIDATE_REPO_PROFILE=repo-core ./scripts/validate-repo.sh` passes, including the mutation lane at 33 of 33. `GOOS=linux go build ./internal/rootio` passes, so the Linux shim compiles. `./scripts/check-hardened-fs.sh`, `./scripts/check-generated-artifacts.sh`, `./scripts/check-doc-language.sh`, `./scripts/check-shell-portability.sh` and `./scripts/check-validator-anchoring.sh` pass. `./scripts/check-pr-shape.sh`: files=10 lines=693 areas=3.

omkhar:
@codex review

omkhar:
Review loop closed clean.

`Codex Review: Didn't find any major issues.` — **Reviewed commit:** `69a5ecbeaa`, which is a prefix of the current head `69a5ecbeaa25ae3e8736a3ffb187075079f52e19`, so the commit gate matches and no push has invalidated it.

Three rounds, three findings, all accepted and fixed, none rebutted, all reacted, all three threads resolved. No unresolved Codex threads remain.

omkhar:
CI status: every required check passes. `GitHub Actions security analysis` failed once on four consecutive HTTP 500 responses from the GitHub API and passed on rerun; the log shows only `curl: (22) The requested URL returned error: 500`, with no finding.

The `Hostile environment` axes `tmpdir`, `uidmap` and `root` fail on `origin/main` today and are unrelated to this change; see the evidence on pull request 704.

PR #725, opened 2026-09-09
Quote the textconv fixture path in the session-commands scenario

omkhar:
The tmpdir hostile-environment axis roots the scenario workspace under a TMPDIR that contains a space, a literal $HOME, and a backtick command substitution. PR #724 fixed the git external clean filter for this axis but left the sibling diff textconv fixture with the same defect, so the tmpdir axis still failed on shared/session-commands.

Git runs a textconv command through a shell. The scenario registered diff.workcell.textconv with an unquoted absolute path and wrote textconv.sh with a double-quoted marker path. Under the hostile TMPDIR the shell word-splits the path and runs the backtick as a command, so the textconv invocation fails.

This quotes both sites, mirroring the filter fix in #724: the textconv.sh marker touch now uses single quotes, and the diff.workcell.textconv value stores a single-quoted path. Scenario intent is unchanged.

Verified: bash -n and shellcheck clean. The full scenario runs only in the container tmpdir axis; the other three axes already pass.


omkhar:
@codex review

PR #727, opened 2026-09-09
Harden two transient CI build and test flakes

omkhar:
Both issues below are pre-existing, transient CI flakes (they cleared on rerun), not regressions. This lane is about to become a required merge gate, so both are worth hardening now.

1. The Go tarball download in the container-smoke build's apt-broker-builder stage (runtime/container/Dockerfile) used a node one-liner fetch with no retry. A single transient network blip failed the whole image build. It now retries up to 5 times with backoff, cleaning the partial download between attempts, mirroring the retry-loop shape the Debian-snapshot fetch already uses elsewhere in the same file. The sha256 integrity check still runs only after a completed download, and only after a final failed attempt does the build fail hard. Versions, SHAs, and URLs are unchanged. internal/metadatautil/pinnedinputs_docker.go carries a literal copy of this stage for pin review, so it moved in lockstep.

2. TestVerifyReleaseOutputsRejectsSymlinkedAssetsDir intermittently failed with ETXTBSY ("text file busy") when exec'ing a freshly written driver script. The race window was that a fixture script's exec bit was set at file-creation time (mode 0o755) rather than after the write was fully closed, so the executable bit and an open writable file description could momentarily coexist on the inode. internal/testkit's shared writeExecutable helper now writes at 0o644, lets the writer close, then chmods to 0o755, guaranteeing the write closes before the file becomes executable and before it is exec'd. Four call sites in internal/testkit/release_outputs_verify_test.go that built their own driver/stub scripts inline now route through this helper instead of duplicating the old write-with-mode pattern.

### DIFF
--- a/internal/metadatautil/pinnedinputs_docker.go
+++ b/internal/metadatautil/pinnedinputs_docker.go
@@ -491,7 +491,20 @@ RUN case "${BUILDARCH}" in \
       *) echo "Unsupported Go build architecture: ${BUILDARCH}" >&2; exit 1 ;; \
     esac \
   && case "${TARGETARCH}" in amd64|arm64) ;; *) echo "Unsupported Go target architecture: ${TARGETARCH}" >&2; exit 1 ;; esac \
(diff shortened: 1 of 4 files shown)


omkhar:
@codex review

